PR #11, opened 2015-09-20
Add an argument for user-specified header fields

jennybc:

BTW I made the error message a bit more informative.

I also changed the capitalization of `Accept` and `User-Agent`. I know header field names aren't *supposed* to be case sensitive and yet, when specifying custom media type, I have seen sometimes they are? The first error here is expected because the custom media type is incorrectly specified in the default headers. But the second error seems to come from specificying the `"accept"` field instead of "`Accept`". The third call demonstrates what this PR is adding.

``` r
gh("/licenses")
 Error: GitHub API error 415 Unsupported Media Type
gh("/licenses",
 Error: GitHub API error 415 Unsupported Media Type
gh("/licenses",
#> [
#>   {
#>     "key": "gpl-2.0",
#>     "name": "GNU General Public License v2.0",
#>     "url": "https://api.github.com/licenses/gpl-2.0",
#>     "featured": true
#>   },
#>   {
#>     "key": "mit",
#>     "name": "MIT License",
#>     "url": "https://api.github.com/licenses/mit",
#>     "featured": true
#>   },
#>   AND SO ON AND SO FORTH
#> ]
```

I guess that's GitHub problem? ¯\_(ツ)_/¯. Anyway, it inspired me to switch the default names to the most common casing.

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -13,24 +13,25 @@ api_url <- "https://api.github.com"
 
 ## Headers to send with each API request
 
-send_headers <- c("accept" = "application/vnd.github.v3+json",
-                  "user-agent" = "https://github.com/gaborcsardi/whoami")
+send_headers <- c("Accept" = "application/vnd.github.v3+json",
+                  "User-Agent" = "https://github.com/gaborcsardi/whoami")
 
 #' Query the GitHub API
 #'
 #' This is an extremely minimal client. You need to know the API
 #' to be able to use this client. All this function does is
 #' \itemize{
-#'   \item Tries to substitute each listed parameter into
+#'   \item Try to substitute each listed parameter into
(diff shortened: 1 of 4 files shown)


jennybc:
This doesn't interact so elegantly with the pagination approach. It would be nicer if the headers from the initial `gh()` call simply propagated, but that's not what current design supports, since only the *content* is available to work with. Not sure what you will think.

jennybc:
Now I see that the header info is present in the attribute. I will take another crack at this, making it play more nicely with pagination.

gaborcsardi:
Yeah, I guess it would be nice

1. automatically add required headers based on the API point. I know, this is now always possible.
2. use the same headers in the pagination.

Again, thanks for the prs, and will look at them properly from next week.

PR #16, opened 2015-09-26
Add an argument for user-specified header fields

jennybc:

I also changed the capitalization of `Accept` and `User-Agent`. I know header field names aren't *supposed* to be case sensitive and yet, when specifying custom media type, I have seen sometimes they are? The first error here is expected because the custom media type is incorrectly specified in the default headers. But the second error seems to come from specificying the `"accept"` field instead of "`Accept`". The third call demonstrates what this PR is adding. The fourth and fifth show that custom headers are used during page traversal.

``` r
gh("/licenses")
#> condition in gh("/licenses"): GitHub API error 415 Unsupported Media Type

## "accept" does not seem to work?!?
gh("/licenses",
#> condition in gh("/licenses", .send_headers = c(accept = "application/vnd.github.drax-preview+json")): GitHub API error 415 Unsupported Media Type

## but "Accept" does
gh("/licenses",
#> [
#>   {
#>     "key": "gpl-3.0",
#>     "name": "GNU General Public License v3.0",
#>     "url": "https://api.github.com/licenses/gpl-3.0",
#>     "featured": false
#>   },
#>  AND SO ON AND SO FORTH
#> ]

## custom headers also work with pagination
two_pages <- gh("/licenses", per_page = 3, .limit = 6,
                .send_headers = c(
                  "Accept" = "application/vnd.github.drax-preview+json"))
str(two_pages, max.level = 1, give.attr = FALSE)
#> List of 6
#>  $ :List of 4
#>  $ :List of 4
#>  $ :List of 4
#>  $ :List of 4
#>  $ :List of 4
#>  $ :List of 4

third_page <- gh_next(two_pages)
str(third_page, max.level = 1, give.attr = FALSE)
#> List of 3
#>  $ :List of 4
#>  $ :List of 4
#>  $ :List of 4
```

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -178,6 +178,7 @@ gh_url <- function(method, url, auth, headers, params) {
 
   attr(res, "method") <- method
   attr(res, "response") <- headers(response)
+  attr(res, ".send_headers") <- headers
   class(res) <- "gh_response"
   res
 }
--- a/R/package.R
+++ b/R/package.R
@@ -13,24 +13,25 @@ api_url <- "https://api.github.com"
 
 ## Headers to send with each API request
 
-send_headers <- c("accept" = "application/vnd.github.v3+json",
-                  "user-agent" = "https://github.com/gaborcsardi/whoami")
+send_headers <- c("Accept" = "application/vnd.github.v3+json",
+                  "User-Agent" = "https://github.com/gaborcsardi/whoami")
 
 #' Query the GitHub API
 #'
 #' This is an extremely minimal client. You need to know the API
 #' to be able to use this client. All this function does is
 #' \itemize{
-#'   \item Tries to substitute each listed parameter into
+#'   \item Try to substitute each listed parameter into
 #'     \code{endpoint}, using the \code{:parameter} notation.
-#'   \item If a GET request (the default), then adds
+#'   \item If a GET request (the default), then add
 #'     all other listed parameters as query parameters.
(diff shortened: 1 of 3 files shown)


jennybc:
Sorry I deleted that branch by mistake. Hopefully I've restored it and reopen this?

jennybc:
ping?

I'm using a branch in my own fork with all 3 of these fixes and it makes me very happy. Specifically, it all works :)

gaborcsardi:
:) Sounds good.

gaborcsardi:
I am still a mess, sorry.... soon.

gaborcsardi:
I think we could just create a table with the api point and required content types, and add the headers automatically. Assuming GitHub does not change them later.... 

But  `.send_headers` is still useful, and we should definitely have it. 

So i'll be happy to merge, but merging the other two messed up the HEAD. I guess. Could you do a rebase? Thanks.

jennybc:
I don't think that table would be very easy to create or maintain. It does seem like this info should exist in a machine readable form yet that doesn't appear to be the case?

These custom media types are coming and going all the time as features go into and out of developer preview. Since I started using `gh` a couple weeks ago, for example, the [new Org permission stuff](https://github.com/blog/2064-new-organization-permissions-now-available) has finally gone live, meaning you no longer need to specify a custom media type for certain requests.

Regardless, yes, later today I'll rebase and make a new PR with this `.send_headers` stuff in it.

PR #17, opened 2015-10-16
Add send headers

jennybc:
*the rebased version of PR #16*

I also changed the capitalization of `Accept` and `User-Agent`. I know header field names aren't *supposed* to be case sensitive and yet, when specifying custom media type, I have seen sometimes they are? The first error here is expected because the custom media type is incorrectly specified in the default headers. But the second error seems to come from specificying the `"accept"` field instead of "`Accept`". The third call demonstrates what this PR is adding. The fourth and fifth show that custom headers are used during page traversal.

``` r
gh("/licenses")
#> condition in gh("/licenses"): GitHub API error 415 Unsupported Media Type

## "accept" does not seem to work?!?
gh("/licenses",
#> condition in gh("/licenses", .send_headers = c(accept = "application/vnd.github.drax-preview+json")): GitHub API error 415 Unsupported Media Type

## but "Accept" does
gh("/licenses",
#> [
#>   {
#>     "key": "gpl-3.0",
#>     "name": "GNU General Public License v3.0",
#>     "url": "https://api.github.com/licenses/gpl-3.0",
#>     "featured": false
#>   },
#>  AND SO ON AND SO FORTH
#> ]

## custom headers also work with pagination
two_pages <- gh("/licenses", per_page = 3, .limit = 6,
                .send_headers = c(
                  "Accept" = "application/vnd.github.drax-preview+json"))
str(two_pages, max.level = 1, give.attr = FALSE)
#> List of 6
#>  $ :List of 4
#>  $ :List of 4
#>  $ :List of 4
#>  $ :List of 4
#>  $ :List of 4
#>  $ :List of 4

third_page <- gh_next(two_pages)
str(third_page, max.level = 1, give.attr = FALSE)
#> List of 3
#>  $ :List of 4
#>  $ :List of 4
#>  $ :List of 4
```

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -13,24 +13,25 @@ api_url <- "https://api.github.com"
 
 ## Headers to send with each API request
 
-send_headers <- c("accept" = "application/vnd.github.v3+json",
-                  "user-agent" = "https://github.com/gaborcsardi/whoami")
+send_headers <- c("Accept" = "application/vnd.github.v3+json",
+                  "User-Agent" = "https://github.com/gaborcsardi/whoami")
 
 #' Query the GitHub API
 #'
 #' This is an extremely minimal client. You need to know the API
 #' to be able to use this client. All this function does is
 #' \itemize{
-#'   \item Tries to substitute each listed parameter into
+#'   \item Try to substitute each listed parameter into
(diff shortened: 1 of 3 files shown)


gaborcsardi:
Thanks much!

PR #18, opened 2015-10-19
attempt endpoint parameter substitution with exactly 1 item

jennybc:
Silences a warning that occurs whenever user supplies input that has length greater than one. This sort of input is ultimately destined to become a query parameter and not substituted into the endpoint URL. I don't think this needs to be documented (?).

Example: the `labels` one can provide when opening an issue.

``` r
library(gh)

retval <- gh("POST /repos/:owner/:repo/issues",
             owner = "STAT545-UBC", repo = "test-repo",
             title = "hw04 peer review", body = "review your peer's work",
             assignee = "jennybc", labels = c("peer-review", "hw04"))
#> Warning in gsub(paste0(":", n, "\\b"), p, endpoint): argument 'replacement'
#> has length > 1 and only the first element will be used
```

### DIFF
--- a/R/parse_end_point.R
+++ b/R/parse_end_point.R
@@ -6,7 +6,7 @@ parse_endpoint <- function(endpoint, params) {
   done <- logical(length(params))
   for (i in seq_along(params)) {
     n <- names(params)[i]
-    p <- params[[i]]
+    p <- params[[i]][1]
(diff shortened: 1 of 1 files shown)


jennybc:
I don't think the AppVeyor failure could possibly be due to this tiny change.

gaborcsardi:
Yes, this is reasonable. Thanks!

Issue #19, opened 2015-10-22
Github Enteprise: allowing users to specify custom Github API url?


akhmed1:
Hello everyone! Github API url is currently hard coded inside `package.R`
```r
api_url <- "https://api.github.com"
```
which would require re-building the library for Github Enterprise users.

Would it make sense to add the API url as an extra parameter to `gh` just as `.token` is? For example:
```r
gh <- function(endpoint, ..., .token = Sys.getenv('GITHUB_TOKEN'),
                .api_url = Sys.getenv('GITHUB_API_URL'),
               .limit = NULL, .send_headers = NULL) { ... }
```

gaborcsardi:
Yes, I think something like this would make sense:

```r
default_api_url <- "https://api.github.com"
...
gh <- function(endpoint, ..., .token = Sys.getenv("GITHUB_TOKEN"),
                .api_url = Sys.getenv("GITHUB_API_URL", unset = default_api_url),
                .limit = NULL, .send_headers = NULL) { ... }
```

PR #20, opened 2015-10-22
Added Github Enterprise support

akhmed1:
Added support for Github Enteprises: allowing users to specify custom Github API url as discussed in #19

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -9,7 +9,7 @@ NULL
 
 ## Main API URL
 
-api_url <- "https://api.github.com"
+default_api_url <- "https://api.github.com"
 
 ## Headers to send with each API request
 
@@ -35,6 +35,7 @@ send_headers <- c("Accept" = "application/vnd.github.v3+json",
 #' @param endpoint GitHub API endpoint. See examples below.
 #' @param ... Additional parameters
 #' @param .token Authentication token.
+#' @param .api_url Github API url (default: \url{https://api.github.com})
 #' @param .limit Number of records to return. This can be used
 #'   instead of manual pagination. By default it is \code{NULL},
 #'   which means that the defaults of the GitHub API are used.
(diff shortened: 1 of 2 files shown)


R/package.R
```diff
@@ -81,9 +82,14 @@ send_headers <- c("Accept" = "application/vnd.github.v3+json",
 #' gh("/licenses") # error code 415
 #' gh("/licenses",
 #'    .send_headers = c("Accept" = "application/vnd.github.drax-preview+json"))
+#'  
+#' ## Access Github Enterprise API
+#' gh("/user/repos", type = "public", .api_url = "https://github.foobar.edu/api/v3")
```
gaborcsardi: You need to put this in `\dontrun{}`, otherwise running the examples fails.

R/package.R
```diff
@@ -35,6 +35,7 @@ send_headers <- c("Accept" = "application/vnd.github.v3+json",
 #' @param endpoint GitHub API endpoint. See examples below.
 #' @param ... Additional parameters
 #' @param .token Authentication token.
+#' @param .api_url Github API url (default: \url{https://api.github.com})
```
gaborcsardi: Please also document the `GITHUB_API_URL` environment variable.

R/package.R
akhmed1: I think it is already inside the existing `\dontrun{}`. (See the new line 88 below -- the closing bracket)

R/package.R
akhmed1: Sure!  `GITHUB_TOKEN` is documented only in Readme. Would that be a good place for `GITHUB_API_URL` too?

R/package.R
gaborcsardi: Sorry, my bad.

R/package.R
gaborcsardi: No, the README is not necessarily comprehensive. They should both be here. But I can add them together, no worries.

gaborcsardi: No, the README is not necessarily comprehensive. They should both be here. But I can add them together.

gaborcsardi:
Thanks!

akhmed1:
Awesome, thanks!

PR #25, opened 2015-11-27
enforce .limit; fixes #24

### LINKED ISSUE #24: .limit only in force if greater than current page size
Right now `.limit` only comes into force if it is greater than current page size:

``` r
library(gh)
iss_list <- gh("/repos/hadley/dplyr/issues", .limit = 4)
length(iss_list)
#> [1] 30
```

Would you rather change documentation or actually limit the returned output to the first `.limit` results?

jennybc:


gaborcsardi:
Great, thanks!

PR #31, opened 2016-01-27
retain attributes when .limit is enforced

jennybc:
as per this discussion https://github.com/gaborcsardi/gh/commit/d2f8a7cab5b746c5d4a4599bb4406fd1b272a2e0#commitcomment-15717354

It wasn't just the `gh_response` class that was being lost but all attributes. Is there some more compact way to do this?


gaborcsardi:
I think your solution is quite compact, I don't think there is a better way.

gaborcsardi:
Thanks!

PR #36, opened 2016-09-05
URLencode() the url

jennybc:
Example where this matters: label handling. Because labels can (a) contain whitespace and (b) appear in the endpoint.

Here's a before example (I delve into this lousy error message below):

``` r
gh("/repos/gaborcsardi/gh/labels/:name", name = "help wanted")
#> No encoding supplied: defaulting to UTF-8.
#> Error in res$message : $ operator is invalid for atomic vectors
```

and after the change in this PR:

``` r
gh("/repos/gaborcsardi/gh/labels/:name", name = "help wanted")
#> {
#>   "url": "https://api.github.com/repos/gaborcsardi/gh/labels/help%20wanted",
#>   "name": "help wanted",
#>   "color": "159818"
#> }
```

So what's up with the unhelpful error message? For some reason this triggers an HTML response (screenshot attached) and currently these lines don't anticipate that: https://github.com/gaborcsardi/gh/blob/24991010aefacf49a2c5cc1c476ffe837706370c/R/package.R#L184-L201

screen shot 2016-09-05 at 4 24 47 pm


gaborcsardi:
Looks great! Thanks very much for fixing this!

Issue #35, opened 2016-09-05
Debug mode? Refactor gh_url()? 


jennybc:
I am willing to take a crack at this but would like your approval and some vague direction.

I'm hitting a lot of new endpoints and discovering new edge cases that break `gh`.

It would be very helpful if it were easier to see the endpoint `gh()` arrives at after the `parse_endpoint()` call and to see what's left for the query or body. And then to see what the body looks like after being sent through `jsonlite::toJSON()`.

Are you generally on board with this?

If so, feel free to say anything --even very vague -- about how you would approach it and I'll try to comply.

jennybc:
I also think refactoring that exposed the fully-substituted endpoint and the query / body, prior to making the API call, would be helpful for writing tests.

gaborcsardi:
YES! This would be indeed great to have. 

I am not sure about the best way, so it will be somewhat experimental, but that's ok, you need to start somewhere. :)

Some random ideas:
* logging, i.e. like a verbose mode in a command line app
* callback function(s), that are called as the request is made, the response is received, etc. They can default to log to the screen, but then you can set them to `browser` or whatever you want.
* configuration via `options()`, maybe.

Since this is experimental, I don't think it needs to be fully fleshed out before we start playing with it. Something dirty is perfectly fine, just to see if it is the right direction.

gaborcsardi:
I think we close this now.

PR #37, opened 2016-09-06
some API errors result in a non-JSON response

jennybc:
This seems a rather weird solution, so feel free to close.

I don't know why the GitHub API would return an error (400) but include the "error message" as html, not JSON. Also not put the status code in the headers. And yet it can happen for an ill-formed endpoint. If I try to get the "help wanted" label (prior to applying PR #36), you get such a mysterious failure. 

This PR gives the user slightly more useful information for troubleshooting:

``` r
gh("/repos/gaborcsardi/gh/labels/:name", name = "help wanted")
#> Error in gh("/repos/gaborcsardi/gh/labels/:name", name = "help wanted") : 
#>   GitHub API error (400):
#>   html response :(
```

and then this shows up in user's browser:

screen shot 2016-09-05 at 4 24 47 pm

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -186,8 +186,13 @@ gh_url <- function(method, url, auth, headers, params) {
   } else if (grepl("^application/json", heads$`content-type`,
                    ignore.case = TRUE)) {
     res <- fromJSON(content(response, as = "text"), simplifyVector = FALSE)
+  } else if (grepl("^text/html", heads$`content-type`, ignore.case = TRUE)) {
+    x <- tempfile(pattern = "gh-github-api-error-", fileext = ".html")
+    write(res, x)
+    browseURL(x)
+    res <- list(message = "html response :(")
   } else {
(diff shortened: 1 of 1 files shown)


gaborcsardi:
I think the API responds with HTML, because `gh` sends a request with invalid syntax. Ideally `gh` should not do that, and whatever the user specifies, it should send  a syntactically valid request.

Still, I think it is good idea to anticipate `gh` errors, so I like this PR. :) I am not sure if we need to open a browser, though. I think that we could just say that this is probably a bug in `gh`.

jennybc:
OK I will pare this down and tag you again. I'm happy to start writing tests for these edge cases as I come across them (#35). I have more 😢.

jennybc:
I'm going to close this. I will take a few things apart to work on #35, which will result in a dedicated function for checking and processing the response. I'll revisit this there and take your comments above into consideration.

PR #38, opened 2016-09-12
Refactor request prep, execution, processing; relates to #35

jennybc:
This is a monster PR 👹, sorry. But it really wasn't possible to do this in small steps. Consider this step one of the experiment discussed in #35.

I re-organized the flow as:

  * Build the request with `gh_build_request()`, which sequentially calls helper functions to process user input into pieces of the request. You will recognize lots of the logic.
  * Make the request with `gh_make_request()`.
  * Post-process the response with `gh_process_response()`. You will also recognize this logic.

If this looks promising, I will start adding tests. I have successfully run all the examples and everything in the README and quite a few of my own scripts, including calls to github.com and to a private GHE instance. I feel like I have basically put everything back together again. Though I'm sure the tests will be educational!

I made a few other changes that were floating around in my local version. I comment on those below inline.

Now you can call `gh_build_request()` (not yet exported or documented, but probably should be) to see something very close to what will go out:

``` r
> gh_build_request("POST /orgs/:org/repos",
+                  params = list(org = "ORGNAME",
+                                name = "REPONAME", description = "DESCRIPTION",
+                                private = TRUE, teamid = 1234))
$method
[1] "POST"

$url
[1] "https://api.github.com/orgs/ORGNAME/repos"

$headers
                                          Accept 
                "application/vnd.github.v3+json" 
                                      User-Agent 
         "https://github.com/gaborcsardi/whoami" 
                                   Authorization 
"token XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"

$query
NULL

$body
{"name":"REPONAME","description":"DESCRIPTION","private":true,"teamid":1234} 
```


R/gh_response.R
```diff
@@ -0,0 +1,33 @@
...
+  stopifnot(inherits(response, "response"))
+  heads <- headers(response)
+
+  content_type <- http_type(response)
+  if (is.null(content_type) || length(content_type) == 0) {
+    res <- ""
+  } else if (grepl("^application/json", content_type, ignore.case = TRUE)) {
+    res <- fromJSON(content(response, as = "text"), simplifyVector = FALSE)
+  } else {
+    if (grepl("^text/html", content_type, ignore.case = TRUE)) {
```
jennybc: This is the new scaled back handling of an html response (#37). I just throw a warning about html content and put it in the message, in case user wants to deal with it.

R/package.R
```diff
@@ -91,27 +82,18 @@ send_headers <- c("Accept" = "application/vnd.github.v3+json",
...
-  url <- URLencode(paste0(.api_url, endpoint))
+  raw <- gh_make_request(req)
 
-  res <- gh_url(method, url, auth, headers, params)
+  res <- gh_process_response(raw)
 
-  while (! is.null(.limit) && length(res) < .limit && gh_has_next(res)) {
-    res2 <- gh_next(res, .token = .token)
+  while (!is.null(.limit) && length(res) < .limit && gh_has_next(res)) {
+    res2 <- gh_next(res)
```
jennybc: The token is already in the response, as part of the headers. So I just re-use the headers for calls for other pages. It doesn't seem like the token needs to be an argument? Or am I missing something?

R/pagination.R
```diff
@@ -5,9 +5,9 @@ extract_link <- function(gh_response, link) {
   if (is.null(links)) {
     return(NA_character_)
   }
-  links <- trim_ws(strsplit(links, ",")[[1]])
+  links <- trimws(strsplit(links, ",")[[1]])
```
gaborcsardi: I would stay with our own `trim_ws`. R's `trimws` is slightly broken, it does not handle all unicode space characters. It would also require R >= 3.2.0.

gaborcsardi:
Looks great! Thanks much! 
Do you want me to merge it now or do you want to add tests first?

R/pagination.R
jennybc: I reverted that.

jennybc:
I'm glad you are willing to merge it! It's encouraging.

Maybe you should give it several days. I will add some tests and have plenty of opportunities to use it this week.

Two questions:

  * Should `gh_build_request()`, `gh_make_request()`, `gh_process_response()` be exported? I suspect so. 
  * Do you have a policy on testing? As in, only test the API exposed to the user? Or should I write tests for some of the small internal functions? Such as the one that substitutes parameters in the endpoint.

jennybc:
Also, do you understand why the appveyor build fails? As the tests actually appear, it would be gratifying if that actually worked.

gaborcsardi:
> Maybe you should give it several days. I will add some tests and have plenty of opportunities to use it this week.

OK, I am travelling for the rest of the week, anyway.

> Should gh_build_request(), gh_make_request(), gh_process_response() be exported? I suspect so.

I would not export them, actually. Why would the user need them? For debugging they can `trace()` or call the internal methods. Let's wait until somebody actually needs them exported.

> Do you have a policy on testing? As in, only test the API exposed to the user? Or should I write tests for some of the small internal functions? Such as the one that substitutes parameters in the endpoint.

I would test every sufficiently difficult function, internals as well. Rule of thumb: if you cannot immediately see that a function is correct, it should be tested. However, I don't think we need to write tests until we are happy with the API.

> Also, do you understand why the appveyor build fails? As the tests appear, it would be gratifying if that actually worked.

CRAN was broken. It should be OK now.

jennybc:
@gaborcsardi 

I pushed a couple more commits, based on heavy usage this week. I haven't added any more tests yet but can say that I have used this ALOT. With github.com and GHE. To create teams, create repos, push files into repos, add users to teams, give teams various permission w/r/t repos, etc.

I'm sure it's not perfect but I think the PR still represents progress. I think it is as or more robust than current master.

If you're willing to merge, I say go ahead.

I can start a new branch where I'll add tests and open a new PR.

gaborcsardi:
Thanks much! How are you planning to write the tests, btw?

FWIW I found a comprehensive, machine-readable  description of the API here: https://github.com/mikedeboer/node-github/blob/master/lib/routes.json
They are using it to auto-generate the whole API, but not for testing I think.

At some point I am planning to write a similarly auto-generated GH api package. This won't be `gh`, but sg with proper R6 classes.

jennybc:
> How are you planning to write the tests, btw?

Good question. My tendency would be to hand craft them, using examples and endpoints I am actually using. But I know it should probably be done in some automated way, using something like the API description you've found.

What is the subtext of your question? Do you think having some artisanal tests is useful in the short-term, even if the long-term plan is to ... generate the tests automatically or even scrap the whole thing in favour of an auto-generated package? Or maybe not bother?

jennybc:
Also, next time I make a PR, can I make myself a contributor or something, in DESCRIPTION?

gaborcsardi:
> Also, next time I make a PR, can I make myself a contributor or something, in DESCRIPTION?

Obviously, I am sorry, long time due.

gaborcsardi:
Re tests, I was just curious if you actually want to hit the GH API, or just test putting together the requests, and interpreting the responses. Either way is fine with me, just curious.

In the long term I want to have automatic tests, it's just easier to cover the API. But I would not scrap hand-written ones, anyway. 

But I think it is up to you, really. If you are comfortable with the current state, then all is fine :)

jennybc:
I was probably NOT going to hit the API because I know from googlesheets that it adds a whole lot of complication. I will take advantage of the new structure to test forming requests and post-processing them. And use the techniques from your blog post about mocking!

I think I would still like to add some hand-written tests in the short-term because it makes it so much easier to fiddle with things and be fairly sure I haven't broken things horribly.

PR #42, opened 2016-11-08
Quick first parse at location github remotes

### LINKED ISSUE #40: Determine github associated with current git repo?
I'm not sure if this belongs here (esp. since it adds a dependency on git2r), but it wouldn't be an unreasonable home.

hadley:
Fixes #40 

For commentary on overall structure before I add documentation/tests. Questions:

* Which of these functions should be exported? Just `github_remote()`?
* `github_remote()` isn't a very good name. Any thoughts? 
* What would you like to see tests for?


gaborcsardi:
> Which of these functions should be exported? Just github_remote()?

I think so. This is a `gh` package, not a `git` package.

> github_remote() isn't a very good name. Any thoughts?



> What would you like to see tests for?

I think we can use `rprojroot` to find the git dir, soon: https://github.com/krlmlr/rprojroot/issues/19
So leave that.

gaborcsardi:
Sorry, posted prematurely the previous one, please see again.

gaborcsardi:
Oh, I also closed it apparently. Sorry about that...

gaborcsardi:
`gh_tree_remote` ? Not sure if it is better, at least it refers to the local tree.

Btw. if there are multiple GH remotes, maybe we want to select `origin` first? I often have `origin` and `upstream` for forks, and I would guess this is common.

hadley:
Ok, I think I've handled everything. PTAL.

hadley:
I figure we can rip out the root detection code once it's supported by rprojroot (which can also handle `GIT_CONFIG` and `GIT_DIR`)

gaborcsardi:
Yes, we can do GIT_CONFIG & GIT_DIR later. 

Btw. why not just return all GH repos? Isn't that useful in `devtools`? Or elsewhere? Anyway, I'll merge this and see what we need from it in the future. 

Thanks!

PR #45, opened 2016-11-08
Handle absolute urls

### LINKED ISSUE #27: Handle url "pointers"
I make a call to a github api, and get a url: "https://api.github.com/repos/cloudera/gestalt/subscription". How can I use that url with gh?

hadley:
Fixes #27 

I had to monkey around with the regexps a bit.  I don't think I broke any valid endpoints, but I may have allowed more invalid endpoints to propagate.


gaborcsardi:
Thanks, looks good!

THB I have no idea if it works with enterprise URLs, but we can test that later.

hadley:
@jennybc has access to an enterprise install if we need a tester.

PR #48, opened 2016-11-25
Strip non names

jennybc:
If all names are `""`, they need to be stripped, in order to produce the correct JSON body. Discovered when trying to set issue labels.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -32,7 +32,7 @@ has_no_names <- function(x) all(!has_name(x))
 
 ## if all names are "", strip completely
 cleanse_names <- function(x) {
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -30,6 +30,14 @@ has_name <- function(x) {
 
 has_no_names <- function(x) all(!has_name(x))
 
+## if all names are "", strip completely
+cleanse_names <- function(x) {
+  if (all(has_no_names(x))) {
```
gaborcsardi: I think one of the `all()` calls is not needed. Otherwise looks good.

gaborcsardi:
Thanks much!

Issue #8, opened 2015-06-11
Test cases


gaborcsardi:
Kinda tricky, as I don't want to make a hundred requests for each test run..... but maybe a mixture of mocking and real test cases.

jennybc:
What are you thinking re: tests that require a token and tests that will actually change things on GitHub?

akhmed1:
How about "personalized" tests?

* tests are disabled by default but could be enabled by a specifying an extra option regarding the level of testing.
* tests are run using each developer's own special test token (taken, say, from an environment variable)

I can see three progressive levels of these tests:

1. Requires no tokens
2. Requires a token but can be done with a free Github account (say, with a throwaway account)
3. Requires a token and requires a paid account (working with organizations, private repos etc)

gaborcsardi:
I think the throwaway account is a good way, if the GH policy allows it. It will still be somewhat tricky to perform the tests that DELETE, PUT, etc. in a safe way.

It would make sense to look at how other similar projects are doing the testing. E.g. the Python GH API, the JS one, etc.

jennybc:
I'm not sure if this is the main Python wrapper (?) but it appears PyGithub's tests run as current user:

https://github.com/PyGithub/PyGithub/blob/master/Contributing.rst

According to the README, the library covers almost the full API and "all methods are tested against the real Github site".

It seems like there's two problems: having fixed entities to test against and having permission to do interesting things. They're related but not exactly the same.

Maybe it would make sense to have an organization that makes fixed repos available for testing. And to do full testing, someone would have to either have a paid account w/ sufficient privileges (can create a private repo, for example) or have a role within that organization that carries the necessary permissions.

gaborcsardi:
@jennybc I would not worry about the private repos much at this point.

I don't mind creating a robot user or org to do the testing. I still have two issues:
* Hundreds of API calls for a test suite. Maybe this is not that bad, and we don't need to worry about it too much just yet..
* How not to expose the credentials to the robot org.

gaborcsardi:
I guess the credentials might be just encrypted variables on the CI, so problem solved....

jennybc:
When you say "How not to expose the credentials to the robot org", you mean, like, encrypting the personal access token or OAuth token?

Yes I see what you mean about the slowness. Maybe you set it up so it's fairly easy to make the testing more or less thorough?

gaborcsardi:
So, yes, @akhmed1 's "personal testing" is good, I think. 

[...]

PR #49, opened 2016-11-27
Explore routes.json

jennybc:
This is more for discussion than merging. Related to #8.

I started to poke around that JSON file you found, documenting the GitHub API.

There are just under 400 endpoints there, broken down into sections.

I've made a table ([visit a rendered version](https://github.com/jennybc/gh/blob/api-description/tests/testthat/routes/routes-table.md)) with one row per endpoint and these columns:

  * section (e.g. "issues")
  * method, url, description
  * params, then broken into url substitution params and other params

Were you already working on this?

I don't understand that node wrapper well enough to tell how they use these routes to produce their tests. For example, I don't see where they give all these parameters values. It seems like they do actually place the requests, because I see evidence of token management.


gaborcsardi:
Thanks! 

I only looked at it really briefly. The API and test generation is here: https://github.com/mikedeboer/node-github/blob/eef99a4979edd6053b8b9af38b82709c310a4443/lib/generate.js#L41-L85

It looks like they only test one set of options for each endpoint. Which is fair if you have ~400 endpoints.

I am wondering if we should generate the test cases on the fly? This is simpler to implement, but could be a nightmare to debug if sg goes wrong.

gaborcsardi:
I think their tests are just broken....

To make tests work, we would need to include meaningful parameters in the JSON, or rather another companion JSON. :(

gaborcsardi:
But maybe it is not so bad. I created a user and an org for testing:
https://github.com/gh-testing
https://github.com/gh-testing-org

We could just use these and two default repos by default, and add parameters for the endpoints that have extra parameters, in another json file:
```json
"/repos/:owner/:repo/deployments" : {
  "owner": "foo",
  "repo": "bar"
}
```
to override/extend the defaults. 
  
Creating this once seems OK, but I am not yet convinced that it can be maintained easily....

jennybc:
This seems like a important early question: will we think of routes.json from note-github as a dependency? And try to keep it up-to-date with them? Or take it as a one-time information dump to start something over here?

Seems like there has to be:

  * routes.json or equivalent, documenting known endpoints
  * parameters.json, defining default values for parameters found in the endpoints
  * something else that specifies which endpoints we will actually test
    - individually or via criteria, such as GET requests that can be formed only using parameters for which we've set defaults) and any additional info (such as over-riding or augmenting default parameter values)

Maybe we would't rig this up to actually trigger test failure. "Success" could just mean successfully forming some sort of a table like this for human inspection:

  * one row per endpoint
  * one column for
    - formed request
    - placed request
    - processed request
 
and a code for "skipped/not attempted", "passed", "failed" (also incorporating your new mocking approach). It could report a summary and more details on the failures.

Does this sound generally sensible? If so, I will try to make some progress and check back in. Or again, I can step away if you find it easier to do yourself.

jennybc:
Semi-off-topic: Just having routes.json does mean we could generate documentation for all the endpoints within gh. I can't decide if that would be worth it or not though.

gaborcsardi:
I would treat `routes.json` as a dependency, but copy it over. Then we would update it as we please. If we only use it for tests, this is not terribly important, I mean when and how we update.

I think we can test "all" endpoints that we have parameters for. A simple approach is to check if the endpoint is in `test-params.json`. If yes, we generate a test for it. If not, we don't. So we don't have to write all parameters to begin testing.

I think it is not crucial to have `parameters.json`, we don't need it for the tests.

What's wrong with triggering test failures? I think we can do that, there is nothing wrong with it. We don't have to overcomplicate this. :)

I can create a token for the `gh-testing` user and share that with you using `hadley/secure`, then you can actually run the tests. I'll try to get to it today evening.

As for docs, yes, that would be great, `node-github` has very nice auto-generated docs at http://mikedeboer.github.io/node-github/ But this is for later. It would probably make more sense if we auto-generated the API client as well. I have some ideas for that, actually.

gaborcsardi:
It seems unlikely that we would do this in gh any time soon, so I am going to close this. The routes.json files still exists at a different location, and it is updated regularly, so we might do it later, or in another package: https://github.com/octokit/rest.js/blob/master/plugins/rest-api-endpoints/routes.json

jennybc:
Yes, I think it was past time to put this to rest ⚰️🙂

I would be much better at this sort of thing now!

PR #50, opened 2016-12-05
Add additional classes to returned error objects

jimhester:
This allows you to catch a github_error or the exact status code
directly.

### DIFF
--- a/R/gh_response.R
+++ b/R/gh_response.R
@@ -21,7 +21,7 @@ gh_process_response <- function(response) {
       headers = heads,
       message = paste0("GitHub API error (", status_code(response), "): ",
                        heads$`status`, "\n  ", res$message, "\n")
-    ), class = c("condition", "error"))
+    ), class = c("github_error", status_code(response), "condition", "error"))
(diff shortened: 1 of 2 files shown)


R/gh_response.R
```diff
@@ -21,7 +21,7 @@ gh_process_response <- function(response) {
       headers = heads,
       message = paste0("GitHub API error (", status_code(response), "): ",
                        heads$`status`, "\n  ", res$message, "\n")
-    ), class = c("condition", "error"))
+    ), class = c("github_error", status_code(response), "condition", "error"))
```
gaborcsardi: Hmmm, just the status code looks strange to me. How about `paste0("http_error_", status_code(response))`?

gaborcsardi: Also, can you please switch `condition` and `error`, `error` is more specific and should come first. (I know you didn't do that, but we might as well fix it, too.)

jimhester: Done!

gaborcsardi:
Thanks!

PR #51, opened 2016-12-08
Add a whoami function (user, token, scopes); closes #39

### LINKED ISSUE #39: Export gh_token()?
Or provide some way to prospectively determine if the user has credentials in place.

jennybc:
Are you willing to have such a function here? I think it would be handy for new users and troubleshooting. But I could put it in githug instead, if you'd prefer not.

If you are receptive, I will flesh out the "HERE'S HOW TO GET A TOKEN AND WHERE TO STICK IT" message and add a test. And do anything else you suggest.


gaborcsardi:
Hmmm. I can see that this is very useful, so it is tempting. 

I am also not sure if `gh` is the right place for it. Why do you think it is better placed here than in `githug`? 

TBH, I still haven't given up the hope of having a proper higher level GH API package, and I think `gh_whoami` could go there. But we don't have that package right now, and I won't write it anytime soon, so if you think it is better placed here than in `githug`, I'll be happy to merge it. :)

jennybc:
> Why do you think it is better placed here than in githug?

Short-term, this is much closer to going to CRAN! Even long-term, `gh` should be the least common denominator for GitHub API work from R. I'd recommend anyone doing this in a script or package use `gh`. So it seems like the right place for this sort of diagnostic.

I will finish off the PR today and you can render your final judgment 🙏.

gaborcsardi:
I see `gh` as the basic machinery of a GH API, and `gh_whoami` does not qualify as *basic* to me. :) 

But it will be fine here, for now. (Which, as we know, likely means forever. :)

jennybc:
I also just realized this sort of gets at #39:

> I'm mostly imagining this would be a place to hang the documentation for how to set up your GITHUB_PAT env var, and to give users a way to check that they're correctly configured.

jennybc:
My roxygen seems different from your roxygen.

Testing: shall I assume nothing, i.e. no token necessarily available? Or take advantage of whatever you've rigged up with httrmock (which I have not yet digested)?

jennybc:
BTW I have used successfully with GHE.

jennybc:
From another conversation:

> I can create a token for the gh-testing user and share that with you using hadley/secure, then you can actually run the tests. I'll try to get to it today evening.

I basically see how this works. But yes I need that token. I played around with my own to get this far. I had some very puzzling times until I discovered `httrmock::mocking_status()`! I clearly don't understand the httrmock workflow.

I obviously won't leave the test like this, with my own info in there and skipping on travis and appveyor.

gaborcsardi:
Yes, sorry, I realized that `hadley/secure` is not going to be a long term solution, because it is not going to CRAN, so I started to look for other options. But for now I could just share the token with you. Based on this: https://gist.github.com/jeroenooms/1ffbfbdad40f4aad6657c337a4924f0e I used this code:

```r
github_pubkey <- function(user){
  url <- sprintf("https://api.github.com/users/%s/keys", user)
  keys <- jsonlite::fromJSON(url)
  lapply(keys$key, openssl::read_pubkey)
}

jenny <- github_pubkey('jennybc')
pubkey <- jenny[[1]] # in case there are multiple keys

token <- readLines("~/works/r-pkgs/gh/tests/testthat/github-token.txt")
buf <- openssl::rsa_encrypt(charToRaw(token), pubkey)
cat(openssl::base64_encode(buf))
```

And here is the encrypted token:

```
MLDsFiOAVtdbbORHOCvfbTDMW6k1eQGPu7PfIXqzDe4SagO49eMAsv0f1g03eGORS+QU53MllUDPmYzlYczUVIdnNPjK1yygqS4NJDJNsWsbsDU1I19+KIFDqQ2lkOqq1/3fQeyvtmr7B4FHAbPMAtFLz/OFdbCPbcHurNCtIs9lck3YXp8IZBO9IaKzEEA1M+i+bnWBXxTbU8E0E1PDGPFXTAY34/GLsm7Th2Nz9OZ+EkZC7tLMHFtxuUYd8qi7lOZ6IxMYPs7b6t6QQ0l0QprHOUQWEVe7Ra/2m9Q5OWpvzoHNDkXFBwdSiNVQb0UhUpOfHL3M4yTjaZ/tYIZ3ZgE4nhKsogGLqhdPhuTXsFxbIqH1AxS+1ugOKcZ+f9VDdN2m/tdmjooxUITJI5+bNBSV9iawSaikD6nLPydqLIdH3a3UJ6akcMbBcjw10jvC67Q8JJhHiJVXjcIHKqwPfnVkwpPlAp6Lnn1Y3NNuzQsnIQJeRjQStd8kTxKNBake3X077d4PF7VloBghu396Waudjqmjy5D/VIIs+tvRlMkGDnf45r/MwjFPENIz9vXesNxSvmm7QdBULRJKcbrM62/BgVntEDUy2sF7nTFCaLnzbpRmz2H+msAAkugSyem3GGrkSScWz1iAGIQRrXFvM5AXQJGirVsM2XZdAu+c6ig=
```

You can just decrypt it with your private key.

gaborcsardi:
CC-ing @jimhester, who might be interested in doing this kind of mocking for `devtools` test cases, for GH queries and even downloads.

As for the `httrmock` workflow, I am also just working it out, so it is quite experimental. Right now we have this:
```r
if (file.exists("github-token.txt")) {
  Sys.setenv(GITHUB_TOKEN = readLines("github-token.txt", n = 1))
  Sys.setenv(DEBUGME = "httrmock")
  httrmock::start_recording()
  httrmock::start_replaying()

} else {
  httrmock::stop_recording()
  httrmock::start_replaying()
}
```

which basically means, reading from the back, that if you don't have a token, then you just want to replay the recorded responses. This is how the tests run on CRAN. Everything that was recorded before is just replayed (`start_replaying`) and requests that were not recorded before are performed, but not recorded (`stop_recording`).

If you do have a token, then, 1) everything that was recorded before is replayed (`start_replaying`, so that a test run does not perform _all_ GH requests), and 2) everything that was not recorded before is performed and then recorded (`start_recording`). 

The idea is that most of the time you don't want to perform all GH requests while developing `gh`, only the new ones. While developing, you might want to change the first block, actually. E.g. to re-record everything, you would do `start_recording` and `stop_replaying`. Or, while adding some features or new API points, or tests for them, you might want `stop_recording` and `start_replaying`, so that all new test requests are performed, but the old ones are not.

Does it make sense? I realize that it is a bit messy, and it would be great to improve it. E.g. we could have some interactive mode chooser, that you could use to select the desired behavior for the current session. I would even show it in my R prompt.

jimhester:
I think one additional thing we need to make this nice is the ability to run a teardown file(s) in testthat / devtools. Because ideally you want recording and replaying on when you are running the tests, but off when you are developing. Also https://github.com/hadley/devtools/issues/1202, https://github.com/hadley/devtools/pull/1169 would be useful for the same reasons.

Also in [lookup](https://github.com/jimhester/lookup) I had planned on writing a function with instructions on setting up a Github Token, as that package won't work well without one. I briefly looked into using https://developer.github.com/v3/oauth_authorizations/#create-a-new-authorization to request a token from the user automatically based on basic authentication, but I am not quite clear how the flow works with two factor authentication. Something like that might be useful for GitHug or here as well.

gaborcsardi:
> I think one additional thing we need to make this nice is the ability to run a teardown file(s) in testthat / devtools. 

Exactly. The code I cited is in "setup", i.e. `helper.R`, but there is currently no teardown in testthat.....

> Something like that might be useful for GitHug or here as well.

Maybe in `githug`. :) I think `gh` should only contain the basic infrastructure, ideally.

jennybc:
Success with the token! Thanks.

> ideally you want recording and replaying on when you are running the tests, but off when you are developing

Right now `helper.R` is run every time you `load_all()`, so it's easy to get into a recording/replaying state without really thinking about it. This is how I puzzled myself because I inadvertently recorded some >400s, tinkering with `gh_whoami()` and nonexistent/bad tokens. Then of course they got replayed until I figured out what I had done. I am wondering if the condition for changing the recording/replaying state should be more than just "`github-token.txt` exists".

> Also in lookup I had planned on writing a function with instructions on setting up a Github Token, as that package won't work well without one. I briefly looked into using https://developer.github.com/v3/oauth_authorizations/#create-a-new-authorization to request a token from the user automatically based on basic authentication, but I am not quite clear how the flow works with two factor authentication. Something like that might be useful for GitHug or here as well.

Yes I would also like to help people store a PAT somehow, from one of these packages. However, given the way the Authorizations API works, it almost feels like we could create as much friction as we remove by using it. In [Happy Git](http://happygitwithr.com/github-pat.html), I have instructions to obtain PAT in the browser and offer this code snippet to help store it:

``` r
cat("GITHUB_PAT=8c70...adf2\n",
    file = file.path(normalizePath("~/"), ".Renviron"), append = TRUE)
```

I know that's very low tech 😔.

gaborcsardi:
> Right now helper.R is run every time you load_all(), so it's easy to get into a recording/replaying state without really thinking about it. This is how I puzzled myself because I inadvertently recorded some >400s, tinkering with gh_whoami() and nonexistent/bad tokens. Then of course they got replayed until I figured out what I had done. I am wondering if the condition for changing the recording/replaying state should be more than just "github-token.txt exists".

Agreed. How about controlling the behavior via environment variables? Then we could have a default behavior, for the user, and developers could change it via a simple command that would just get/set an environment variable?

tests/testthat/test-whoami.R
```diff
@@ -0,0 +1,25 @@
...
+  expect_equivalent(
+    res,
+    list(name = "Jennifer (Jenny) Bryan",
+         login = "jennybc",
+         html_url = "https://github.com/jennybc",
+         scopes = "admin:org, admin:public_key, admin:repo_hook, delete_repo, gist, notifications, repo, user"))
+})
+
+test_that("whoami works in absence of PAT", {
+  expect_message(res <- gh_whoami(.token = ""),
```
gaborcsardi: I think would be great if `expect_error` retained the error object, and we could test the error class. See https://github.com/hadley/testthat/issues/530

jimhester: Well I was using it in lookup https://github.com/jimhester/lookup/blob/e841d72819e39242e0987aa6f23b240c9d47d60c/R/rcpp.R#L3.

But I think if you are expecting a response to have a specific error class you should catch just that specific class and let any other error be signaled normally.

tests/testthat/test-whoami.R
```diff
@@ -0,0 +1,25 @@
...
+                 "No personal access token \\(PAT\\) available.")
+  expect_null(res)
+})
+
+test_that("whoami errors with bad PAT", {
+  expect_error(res <- gh_whoami(.token = NA), "Requires authentication")
```
jimhester: Yes here I would definitely check the error class rather than the message, the message could potentially be changed by GitHub any time, the HTTP error code should be more stable.

jennybc:
> How about controlling the behavior via environment variables? 

That sounds great! Does the environment variable `R_TESTS` help us at all? It will be non-`NULL` when the tests are running via `devtools::test()`.

gaborcsardi:
Yes, R_TESTS could indeed help, but I am a bit reluctant to use it, as it causes trouble for many packages, and some people (including myself) often unset it.

jimhester:
Re: environment variables @lionel- has a PR that should help some https://github.com/hadley/devtools/pull/1391, but it is not yet merged.

tests/testthat/test-whoami.R
jennybc: I am now catching and expecting a specific error class and HTTP error.

tests/testthat/test-whoami.R
jennybc: I am now catching and expecting a specific error class and HTTP error.

gaborcsardi:
Btw. if you `skip()` the tests, then I can merge this, and you don't have to wait for the `httrmock` updates.

jennybc:
I have skipped the one that I know must fail for now. The first one should not. Seeing if setting GITHUB_TOKEN to something other than `NULL` fixes that.

jennybc:
VICTORY. I had to set `GITHUB_TOKEN` to some value on Travis and Appveyor. I hope that's OK. It should never be actually used and, if it is, I guess you'd want to know and fail anyway?

gaborcsardi:
Yeah, I think that is fine.

R/gh_request.R
```diff
@@ -119,7 +119,9 @@ gh_token <- function() {
 
 gh_auth <- function(token) {
   auth <- character()
-  if (token != "") auth <- c("Authorization" = paste("token", token))
+  if (isTRUE(token != "")) {
+    auth <- c("Authorization" = paste("token", token))
+  }
```
gaborcsardi: Sorry, can you write this with `if () { ... } else { ... }`? I think it shows the intent better.

R/gh_response.R
```diff
@@ -21,7 +21,7 @@ gh_process_response <- function(response) {
       headers = heads,
       message = paste0("GitHub API error (", status_code(response), "): ",
                        heads$`status`, "\n  ", res$message, "\n")
-    ), class = c("condition", "error"))
+    ), class = c("github_error", paste0("http_error_", status_code(response)), "error", "condition"))
```
gaborcsardi: This was already added by Jim's PR, no? Can you please rebase, if it is easy? If it is not easy, then don't worry about it, it is not worth getting into git trouble for this.....

R/gh_whoami.R
```diff
@@ -0,0 +1,74 @@
...
+#' Reports wallet name, GitHub login, and GitHub URL for the current
+#' authenticated user, the first few and last characters of the token, and the
+#' associated scopes.
+#'
+#' Get a personal access token for the GitHub API from
+#' \url{https://github.com/settings/tokens} and select the scopes necessary for
+#' your planned tasks. The \code{repo} scope, for example, is one many are
+#' likely to need. The token itself is a string of 40 letters and digits. You
+#' can store it any way you like and provide explicitly via the \code{.token}
+#' argument to \code{\link{gh}()}.
```
gaborcsardi: markdown roxygen parsing is on for this project, so feel free to write markdown if you want to. You would need the dev version of roxygen2, though, so if you don't want to deal with that, that's fine, too.

R/gh_whoami.R
```diff
@@ -0,0 +1,74 @@
+#' Info on current GitHub user and token
+#'
+#' Reports wallet name, GitHub login, and GitHub URL for the current
+#' authenticated user, the first few and last characters of the token, and the
+#' associated scopes.
```
gaborcsardi: Hmmm. I am not a security expert, but I would not print out anything from the token. I know that the user has access to it, but I just don't want it to appear in printouts, Rmds, etc.

Let me think about this a bit. I see that it can be useful for beginners, but I am still concerned.

R/gh_whoami.R
```diff
@@ -0,0 +1,74 @@
...
+#' line that looks like this, substituting your PAT:
+#'
+#' \preformatted{
+#' GITHUB_PAT=8c70fd8419398999c9ac5bacf3192882193cadf2
+#' }
+#'
+#' Put a line break at the end! If you’re using an editor that shows line
+#' numbers, there should be (at least) two lines, where the second one is empty.
+#' Restart R for this to take effect. Call \code{gh_whoami()} to confirm
+#' success.
```
gaborcsardi: Wow! You really have a lot of empathy for beginners.  :)

R/gh_whoami.R
```diff
@@ -0,0 +1,74 @@
...
+gh_whoami <- function(.token = NULL, .api_url = NULL, .send_headers = NULL) {
+  .token <- .token %||% gh_token()
+  if (isTRUE(.token == "")) {
+    message("No personal access token (PAT) available.\n",
+            "Obtain a PAT from here:\n",
+            "https://github.com/settings/tokens\n",
+            "For more on what to do with the PAT, see ?gh_whoami.")
+    return(invisible(NULL))
+  }
+  req <- gh_build_request(endpoint = "/user", token = .token,
```
gaborcsardi: I guess if we do not show the token, then you don't need to build a `gh_request` manually, right?

gaborcsardi:
Thanks much, looks awesome! I added some small comments. 

Most important is about showing part of the token on the screen. I am not sure if that's a good idea to be honest, but I might be paranoid.

R/gh_whoami.R
jennybc: I thought showing first 4 and last 4 characters was OK. I do the same over in googlesheets. It is helpful when you have more than one token in your life and there's no obvious way to give them names. But we can kill it or reveal even less of it.

R/gh_whoami.R
jennybc: I may have answered this question a few dozen times.

R/gh_response.R
gaborcsardi: Yes, right, show be fine I think.

R/gh_whoami.R
jennybc: You're right. I can avoid all the manual stuff and simply use `gh()`. Have simplified that.

R/gh_request.R
jennybc: Done.

R/gh_whoami.R
gaborcsardi: I don't know, TBH. Showing 8 characters makes the token 2^(8*4) times less secure. We would show 32 bits of the 160.

gaborcsardi: On the other hand, this is not sg you can "crack", an attacker would need to try each remaining valid key in an API request....

gaborcsardi: How about we only show the first two letters now? Is that still useful for the user?

I'll also ask my security consultant, @jeroenooms. :)

R/gh_whoami.R
jennybc: I haven't done this here. Maybe after the merge I could do it for all exported functions at once?

gaborcsardi: You don't have to do it, just mentioned it for next time. :) I don't think rewriting it it worth the effort.

gaborcsardi:
OK, I will merge this now, and maybe I'll change the token printout slightly.

gaborcsardi:
Thanks much!

R/gh_whoami.R
jennybc: OK 2 characters it is.

gaborcsardi:
@jennybc FYI, I added a simple version of `httrmock` contexts.

jennybc:
Great! I will pursue that. I might be able to resume the `routes.json` activity this week for automating tests and I'm sure this will come up.

gaborcsardi:
Great! I'll still add better request matching. E.g. for POST we should at least use the data....

jennybc:
I will start with GETs anyway, as I'm sure that will be highly educational.

jennybc:
This PR was a good warm-up because I understand the basic mojo of httrmock now but did not before.

PR #54, opened 2016-12-14
Refactor handling of the testing token; closes #53

### LINKED ISSUE #53: Testing token management
The gh package prefers the token in `GITHUB_PAT` over that in `GITHUB_TOKEN`:

``` r
gh_token <- function() {
  token <- Sys.getenv('GITHUB_PAT', "")
  if (token == "") Sys.getenv("GITHUB_TOKEN", "") else token
}
```

But the testing setup enacts the testing account's token via `GITHUB_TOKEN` (only):

``` r
if (file.exists("github-token.txt")) {
  Sys.setenv(GITHUB_TOKEN = readLines("github-token.txt", n = 1))
  ...
}
```

If a developer has no `GITHUB_PAT`, then when the tests run, the testing token is used and all is well. I suspect this is true for @gaborcsardi.

But I have `GITHUB_PAT` defined with my personal PAT, so to ensure the testing token is used I need to  call `gh()` like so:

``` r
gh(..., .token = Sys.getenv("GITHUB_TOKEN"))
```

It seems like we should do one of these things and I think the second is a better idea:

  1. In `helper.R`, set both `GITHUB_TOKEN` and `GITHUB_PAT` to the testing token.
  1. Agree to specify the testing token explicitly in `gh()` calls in the tests.

What if we agree to store the testing token in its own env var? Then we can all define it in our usual way (e.g. in `.Renviron`). We can have a helper that retrieves it. Existence of this env var becomes the condition in `helper.R` instead of the existence of `github-token.txt`. And all `gh()` calls in tests use the helper to explicitly use the testing token.

I think this might be a nicer development workflow.

`helper.R` would look more like this:

``` r
tt <- function() Sys.getenv("GH_TESTING_TOKEN", NA)

...

if (!is.na(Sys.getenv("GH_TESTING_TOKEN")) {
  ## the mocking stuff still goes here
}
```

And a `gh()` call in the typical test looks like so:

``` r
gh(..., .token = tt())
```

Side benefit: if you ever wanted to make HTTP calls from travis or appveyor, now you just need to encrypt an env var vs a file.

If you approve @gaborcsardi, I will make the PR.

jennybc:
As discussed in #53. I also cleared all recordings and re-recorded them. I'm pretty sure some had been recorded with a token other than the testing token, but it wasn't hurting anything. I also started to use a name convention to flag files that make API requests.


jennybc:
This CONTRIBUTING might just be weird? It's 50% a note-to-self, but 50% a barebones orientation for someone who just happens to drop by. I can delete if you wish.

CONTRIBUTING.md
```diff
@@ -0,0 +1,27 @@
...
+If a token is needed, use the `tt()` helper function to use the testing token:
+``` r
+test_that("some other new thing that requires a PAT works", {
+  res <- gh(..., .token = tt())        
+  expect_that(...)
+  ...
+})
+``` 
+
+The testing token should be stored in the `GH_TESTING` environment variable. New functionality and associated tests can be tested locally by temporarily setting `GH_TESTING` to a personal PAT. When a pull request gets merged, the maintainer can run and record the tests using the private gh testing token and commit them to the repo.
```
gaborcsardi: I don't mind a CONTRIBUTING file, although I can't promise to keep it up to date while things are changing.

gaborcsardi:
Thanks much!

PR #58, opened 2017-02-08
Add an example of forcing input to go as JSON array

jennybc:
I keep needing this and re-deriving it. Maybe this cycle will be faster if I put an example in the package!


gaborcsardi:
Thanks!

PR #63, opened 2017-07-12
Catch ping_port() error when cannot resolve host; fixes #62

### LINKED ISSUE #62: skip_if_offline() errors when cannot resolve host
If I'm offline, `skip_if_offline()` throws an error. Example:

```
1. Error: errors return a github_error object (@test-mock-error.R#5) -----------
Cannot resolve host name
```

jennybc:
Re: #62 `skip_if_offline()` errors when cannot resolve host

### DIFF
--- a/tests/testthat/helper-offline.R
+++ b/tests/testthat/helper-offline.R
@@ -1,6 +1,8 @@
 
 skip_if_offline <- function() {
-  if (is.na(pingr::ping_port("github.com", count = 1, timeout = 0.2))) {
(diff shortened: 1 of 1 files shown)


gaborcsardi:
Thanks! 

I actually tried this, and I only got a harmless message. Interesting.

jennybc:
Hmm ... why do you think I get an error then?

gaborcsardi:
I don't know. Yours definitely is an error. I have no clue why it was different for me. Anyway, your solution is better, so we are fine now.

PR #64, opened 2017-07-25
Common offline status for all tests

jennybc:
I started to use this skipping approach over in googledrive and we are giving it a thorough workout. We find that `skip_if_offline()` can happen sporadically for no genuine reason. I suspect it's better to determine online vs offline once, at setup, then use that state throughout.


gaborcsardi:
Thanks! I think that the time limit that is allowed for the check is too small, hence the random behavior. 

Nevertheless, the PR makes a lot of sense.

PR #72, opened 2018-02-13
Expose more information on client errors

### LINKED ISSUE #60: better error messages
https://github.com/jimhester/lookup/issues/15#issuecomment-305891267

jennybc:
Fixes #60

https://developer.github.com/v3/#client-errors

---

``` r
devtools::load_all("~/rrr/gh")
#> Loading gh

## try to create a label that exists, with an invalid color
gh(
  endpoint = "POST /repos/:owner/:repo/labels",
  name = "good first issue", color = "HELLNO",
  owner = "jennybc", repo = "fluffy-otter"
)
#> Error in gh_process_response(raw): 
#> GitHub API error (422): 422 Unprocessable Entity
#> Message: Validation Failed
#> Read more at https://developer.github.com/v3/issues/labels/#create-a-label
#> 
#> Errors:
#>  resource field           code
#>     Label  name already_exists
#>     Label color        invalid

## try to create a repo that already exists
gh("POST /user/repos", name = "fluffy-otter")
#> Error in gh_process_response(raw): 
#> GitHub API error (422): 422 Unprocessable Entity
#> Message: Repository creation failed.
#> Read more at https://developer.github.com/v3/repos/#create
#> 
#> Errors:
#>    resource field   code                             message
#>  Repository  name custom name already exists on this account
```

Created on 2018-02-12 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).


jennybc:
I think the travis failure is because I've made an external PR and don't have access to the encrypted token. And/or a rate limit problem?

```
> test_check("gh")
── 1. Failure: errors return a github_error object (@test-mock-error.R#11)  ────
`e` inherits from `github_error/http_error_403/error/condition` not `http_error_404`.
── 2. Error: can catch a given status directly (@test-mock-error.R#19)  ────────
GitHub API error (403): 403 Forbidden
Message: API rate limit exceeded for 34.233.56.198. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```

R/gh_response.R
```diff
@@ -14,20 +16,52 @@ gh_process_response <- function(response) {
...
+  errors <- res$errors
+  if (!is.null(errors)) {
+    errors <- as.data.frame(do.call(rbind, errors))
+    nms <- c("resource", "field", "code", "message")
+    nms <- nms[nms %in% names(errors)]
+    msg <- append(
+      msg,
+      c("",
+        "Errors:",
+        capture.output(print(errors[nms], row.names = FALSE))
```
gaborcsardi: Is it possible to avoid `capture.output` somehow? It does not play well with debugging, and terminal coloring, so I think it is good practice to avoid it, if possible.

gaborcsardi: It seems difficult to avoid it. There is `format.data.frame()`, but that still returns a data frame.

gaborcsardi:
Thanks!

R/gh_response.R
jennybc: Yes, I tried `format.data.frame()` myself, among other things. I don't know of any other way to get this result for so little work. If it really matters, I can do this another way.

jimhester: I think it would be better to avoid constructing a `data.frame` at all in this case and just use `print(sep = " ", collapse = "\n")` to construct the format directly.

jimhester: Although I just realized this is already merged, so I guess I am too slow!

gaborcsardi: We can still change it, if you think it is better. I really don't like `capture.output`. Although in this case it is mostly harmless.

jennybc: It's a PITA to format nicely because the number and names of the fields varies across endpoints (see my two examples).

gaborcsardi: Yeah, I think it is just fine as it is. At some point we might want to write some real `format.data.frame()` method, though. That returns actual strings.

jennybc: BTW I think it's good to keep capturing examples of error-triggering calls from real life, so we can sanity check this function as it evolves. It would be impossible to write a useful version of this function based only on GitHub's API docs.

PR #73, opened 2018-02-14
Raw GitHub responses returned as raw vectors

noamross:
-  Check GH media type to determine response type
-  Set raw response class as c("gh_response", "raw")
-  Modify print method to be raw-friendly
-  Document the fact that response type changes
-  Add test, NEWS

This makes the function not strictly type-safe but I think returning a raw vector makes sense for a raw response.

Also, new tests are skipped, as the other repo tests are because they are not mocked, but they pass when run.

### DIFF
--- a/R/gh_response.R
+++ b/R/gh_response.R
@@ -5,10 +5,14 @@ gh_process_response <- function(response) {
   }
 
   content_type <- http_type(response)
+  gh_media_type <- headers(response)[["x-github-media-type"]]
+  is_raw <- grepl("param=raw$", gh_media_type, ignore.case = TRUE)
   if (length(content(response)) == 0) {
     res <- ""
   } else if (grepl("^application/json", content_type, ignore.case = TRUE)) {
     res <- fromJSON(content(response, as = "text"), simplifyVector = FALSE)
+  } else if (is_raw) {
+    res <- content(response, as = "raw")
   } else {
     if (grepl("^text/html", content_type, ignore.case = TRUE)) {
       warning("Response came back as html :(", call. = FALSE)
@@ -19,7 +23,11 @@ gh_process_response <- function(response) {
   attr(res, "method") <- response$request$method
   attr(res, "response") <- headers(response)
   attr(res, ".send_headers") <- response$request$headers
-  class(res) <- c("gh_response", "list")
+  if(is_raw) {
+    class(res) <- c("gh_response", "raw")
+  } else {
+    class(res) <- c("gh_response", "list")
+  }
   res
 }
 
--- a/R/gh_response.R
+++ b/R/gh_response.R
@@ -23,7 +23,7 @@ gh_process_response <- function(response) {
   attr(res, "method") <- response$request$method
   attr(res, "response") <- headers(response)
   attr(res, ".send_headers") <- response$request$headers
-  if(is_raw) {
+  if (is_raw) {
     class(res) <- c("gh_response", "raw")
   } else {
     class(res) <- c("gh_response", "list")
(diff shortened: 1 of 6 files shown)


noamross:
I think the test errors are due to the fact the PRs don't have access to the Travis tokens.  However, should my tests be using `gh(..., .token = tt())`?

gaborcsardi:
Thanks! Very good idea! 

> However, should my tests be using gh(..., .token = tt())?

Yes, please!

noamross:
Done. It looks like the CI errors are due to API limits when not using a token, as the PR can't access the env var.

R/gh_response.R
```diff
@@ -19,7 +23,11 @@ gh_process_response <- function(response) {
   attr(res, "method") <- response$request$method
   attr(res, "response") <- headers(response)
   attr(res, ".send_headers") <- response$request$headers
-  class(res) <- c("gh_response", "list")
+  if(is_raw) {
```
gaborcsardi: Can you please put a space after the `if`?

R/print.R
```diff
@@ -4,11 +4,18 @@
...
 #' @return The JSON result.
-#' 
+#'
 #' @importFrom jsonlite prettify toJSON
 #' @export
 #' @method print gh_response
 
 print.gh_response <- function(x, ...) {
-  print(toJSON(unclass(x), pretty = TRUE, auto_unbox = TRUE))
+  if("raw" %in% class(x)) {
```
gaborcsardi: Can you pls put a space after the `if`?

gaborcsardi: And also, just use `inherits(x, "raw")`?

R/print.R
```diff
@@ -4,11 +4,18 @@
...
+    attr(x, c("method")) <- NULL
+    attr(x, c("response")) <- NULL
+    attr(x, ".send_headers") <- NULL
```
gaborcsardi: Maybe we just want `attributes(x) <- NULL`?

tests/testthat/test-mock-repos.R
```diff
@@ -104,3 +104,21 @@ test_that("repos, some basics", {
...
+  skip_if_offline()
+  skip_on_cran()
+  skip("needs mocking")
+
+  res <- gh(
+    "GET /repos/:owner/:repo/contents/:path",
+    owner = "r-lib",
+    repo = "gh",
+    path = "DESCRIPTION",
+    .send_headers = c(Accept = "application/vnd.github.v3.raw"),
```
gaborcsardi: I wonder if we could just always send this header? Is it s problem for the "normal" end points that return JSON? Or is it a good idea to make the user think about the output format?

R/print.R
noamross: If we do this we lose the class, and I thought it would be useful that it still prints the the class is c("gh_response", "raw"), to remind the user of the origin and that they can query the attributes.

tests/testthat/test-mock-repos.R
noamross: I just tested this on an endpoint that doesn't have a raw response, and it returned JSON fine. 
 But the user may not want the raw version when it is available.  The default for file contents and blobs is to return JSON with the file contents as a base64 field.  If you always send the header, it should probably be the JSON one.  The other options available for some endpoints (`diff`, `patch`, `sha`) are all handled fine as text.

R/gh_response.R
noamross: Done

R/print.R
noamross: Done

tests/testthat/test-mock-repos.R
gaborcsardi: Got it.

gaborcsardi:
Thanks much!

PR #77, opened 2018-02-15
Allow responses to be written directly to disk

noamross:
-  Add `destfile=NULL` and `overwrite=FALSE` args
-  Use httr::write_disk() when destfile provided
-  Make print function path-friendly
-  Document
-  Add tests
-  Update to disk


noamross:
Done.  Also fixed overwriting, which wasn't being carried through properly.

gaborcsardi:
Great, thanks!

PR #78, opened 2018-02-21
A simple interface to the GraphQL endpoint

jimhester:
This turned out to be way less code than I thought originally. So little that I am not sure it is worth having a separate function for. So part of the reason I opened this PR was to get feedback on if this is something we want in gh, or if we should just add an example of using `gh::gh()` with this endpoint.


gaborcsardi:
:) Yeah. I had the same conclusion after looking at the graphql api. It is possible to write a DSL for generating the query, but what would be the point, really?

I don't mind. We could merge this, but I agree that a README example, or maybe even a small vignette would serve people better.

jennybc:
Based on yesterday's wrangling, the real work appears to be on the other end 😱

I agree it's worth putting somewhere linkable, so it can be referred to in conversations re: support for v4 API.

hadley:
Implementing pagination would make it more useful?

gaborcsardi:
@hadley Incidentally I just wrote some code that uses gh and GraphQL with pagination. I think a nice interface is to use a callback function. (Well, short of the async package or generators.) Something like this:

```r
walk_gh_versions <- function(callback) {

  q <- '
  {
    organization(login: "cran") {
      repositories(first: 50,
                   orderBy: { field: NAME, direction: ASC },
                   after: <after>) {
        nodes {
          name
          refs(refPrefix: "refs/tags/", last: 100,
               orderBy: { field: TAG_COMMIT_DATE, direction: ASC }) {
            nodes {
              name
            }
          }
        }
        pageInfo {
          endCursor
          hasNextPage
        }
      }
    }
  }'

  after <- "null"
  has_next <- TRUE
  while (has_next) {
    qx <- sub("<after>", after, q)
    res <- gh("POST /graphql", query = qx)
    data <- res$data$organization$repositories
    callback(data)
    has_next <- data$pageInfo$hasNextPage
    if (has_next) after <- sprintf('"%s"', data$pageInfo$endCursor)
  }

  invisible()
}
```

We could automate some of this though.

jimhester:
That is basically the same thing I did for the dashboard, e.g. https://github.com/jimhester/tidyversedashboard/blob/fdc2f73cfdd9e96987b5da5fcbff8dd5e999d1a3/R/issue_progress.R#L1-L40

gaborcsardi:
Thanks!

PR #85, opened 2018-09-04
Fallback PAT to Interactive OAuth

saurfang:
Add `gh_interactive_oauth_token` function which opens the browser and acquires an OAuth token interactively. This can be used as a fallback when the user has not set PAT. The process of acquiring OAuth token is a much easier user experience than creating a PAT. The fallback is only triggered under interactive R session. Furthermore, this simplified user experience is most relevant to Github Enterprise installation where functions such as `devtools::install_github` can be wrapped to install from GHE instead or use GHE as a clean interface to query LDAP username etc.

Unfortunately, this is not unit testable (and reprex can't fake interactive session), here is an illustration of this working:

```r
library(gh)
# clear tokens that are already set
Sys.unsetenv("GITHUB_PAT")
Sys.unsetenv("GITHUB_TOKEN")

gh_whoami()
```

returns

```json
{
  "name": "Forest Fang",
  "login": "saurfang",
  "html_url": "https://github.com/saurfang",
  "scopes": "",
  "token": "3b..."
} 
```

The function and its logic is based on https://github.com/cscheid/rgithub/blob/e98afc3bbf2971a59536fc3ac5ee1ec7c54ead8b/R/github.R#L50


R/gh_request.R
```diff
@@ -133,16 +134,83 @@ gh_set_dest <- function(x) {
...
+#' install_github <- function(...) {
+#'   devtools::install_github(host = "{ghe_api}", auth_token = token$credentials$access_token, ...)
+#' }
+#' }
+gh_interactive_oauth_token <- function(client_id, client_secret, github_url) {
+  if (!interactive()) {
+    stop("OAuth cannot proceed under a non-interactive session!")
+  }
+  if (missing(client_id) && missing(client_secret)) {
+    # TODO: replace with r-lib id and oauth secret
```
saurfang: If this PR were to be merged, author should register a new github app under `r-lib` org

R/gh_request.R
```diff
@@ -133,16 +134,83 @@ gh_set_dest <- function(x) {
...
 #' if no token is available.
 #'
 #' @export
-
 gh_token <- function() {
   token <- Sys.getenv('GITHUB_PAT', "")
-  if (token == "") Sys.getenv("GITHUB_TOKEN", "") else token
+  if (token == "") token <- Sys.getenv("GITHUB_TOKEN", "")
+  if (token == "" && interactive()) {
+    gh_interactive_oauth_token()$credentials$access_token
```
gaborcsardi: The problem with this, is that gh also works fine without any token, or even a GH account, and this forces you to get a token, right?

jennybc: Yeah it would be nice to think about a workflow where user either says "let's do the OAuth token thing" first. Or fails in a non-scary way and is prompted to "do the Oauth token thing". I.e. make this opt-in, but a very easy to opt-in.

saurfang: Ah yes. Good point! Let me look into a more optional/intelligence fallback. Thanks for the feedback.

gaborcsardi:
I will close this for lack of activity. 

@saurfang if you are still interested in updating it, please reopen or submit a new PR. Thanks!

Issue #26, opened 2015-12-17
progress


jennybc:
Is it feasible for `gh` to indicate progress?

gaborcsardi:
When making multiple rounds of queries? I guess a single query is always fast.

jennybc:
I'm making some queries now that retrieve thousands of items and take several minutes. It would be comforting to have some sense of progress.

I guess I actually know from using the API directly how easy (or, annoying, rather) it is to learn the total number of items available.

What do you think of showing progress, perhaps only when `.limit = Inf`?

gaborcsardi:
Yeah, whenever `.limit` is bigger than 30 we can do it I guess. For the `.limit = Inf` case when there is a `rel="last"` link we can do a proper progress bar, without that we can only do a counter that shows how many queries have completed.

gaborcsardi:
I am rewriting https://github.com/gaborcsardi/progress, again. And it will be awesome, of course. Once that is done, I'll get to this.

PR #86, opened 2018-11-24
Interruptible pagination loop

krlmlr:
to return partial results, with a warning, if an interrupt or error occurs.

We seem to get an error from _curl_ instead of an `interrupt` exception. Is there a way to ask _curl_ to not care about interrupts?

I'd like to see a progress bar, too, but this requires parsing the `page=` part of the URL.

How about implementing `[.gh_response()` and `c.gh_response()` ?

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -132,18 +132,28 @@ gh <- function(endpoint, ..., .token = NULL, .destfile = NULL,
 }
 
 gh_fill <- function(res, .limit) {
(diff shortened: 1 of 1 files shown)


gaborcsardi:
Returning partial results on interrupt is uncommon. Interrupt usually just goes back to top level without returning anything. I think we should do that as well.

If you don't want to lose the partial results, then we'll need another mechanism. E.g. store it in the package namespace, and have a function return it.

krlmlr:
A progress bar is discussed in #26.

PR #87, opened 2018-11-24
Point to .limit argument from gh_next()

krlmlr:
Users looking into `gh_next()` probably want the `.limit` argument. Do we want to advertise even more prominently?


gaborcsardi:
Thanks!

PR #88, opened 2018-11-24
Increase per_page if .limit is given

krlmlr:
so that fewer requests are needed. Need to expose argument to avoid collisions with `per_page` given in `...` .


R/package.R
```diff
@@ -36,6 +36,8 @@ NULL
 #' @param ... Name-value pairs giving API parameters. Will be matched
 #'   into \code{url} placeholders, sent as query parameters in \code{GET}
 #'   requests, and in the JSON body of \code{POST} requests.
+#' @param per_page Number of items to return per page. The default varies
+#'   across API endpoints and depending on the `.limit` argument.
```
gaborcsardi: Could you please elaborate? Just say that the default is the minimum of 100 and `.limit`, and the GH API limit.

PR #96, opened 2019-07-12
Document environment variables

### LINKED ISSUE #95: Document GITHUB_API_URL env var for changing default one
Following r-lib/usethis#834, it seems that this feature implement in #20  is not documented. (and now lives [here](https://github.com/r-lib/gh/blob/bed43d354789d2e4dde75db0327da1d46358f953/R/gh_request.R#L108))

I think a sentence about this env var could be added in every function that uses `.api_url`. 

Even something in the readme could be useful. I guess it depends if you plan a pkgdown site for this r-lib tool. Maybe documentation on how to configure a different default GH url should live in usethis. 

I am in the train right now but happy to PR doc improvement anytime.

cderv:
This fixes #95 

* Added a section in README
* Added a few lines in help files and examples.

This should help people find this feature.

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -44,9 +44,11 @@ NULL
 #'   be written to disk in the form sent.
 #' @param .overwrite if \code{destfile} is provided, whether to overwrite an
 #'   existing file.  Defaults to FALSE.
-#' @param .token Authentication token.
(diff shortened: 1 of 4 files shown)


gaborcsardi:
Great, thanks much!

PR #123, opened 2020-07-31
URL-based, credentials-backed token lookup

### LINKED ISSUE #110: Token: check for env var first, then the keyring
That's a better user experience, because the keyring check might bring up some dialogs, even if the token is in an env var.

### LINKED ISSUE #120: Use credentials::git_credential_ask() to summon the token for a given host
This is basically the gh analogue of what gert does for https remotes.

Basic sketch:

If the host is GitHub.com and GITHUB_PAT is set, use it.

Otherwise ask credentials to lookup / request (and then store) the PAT for GitHub.com or whatever GHE instance is implied by the `host`.

jennybc:
WIP

Closes #120 Use credentials::git_credential_ask() to summon the token for a given host
Closes #110 Token: check for env var first, then the keyring

The best place to start your review is probably by reading the new vignette, which (re-)orients you to how things work in gh and explains aspects of credentials that I have wired in. Also raises a few questions and TODOs.

@jeroen This is going to look very familiar, because `set_github_pat2()` here is similar to https://github.com/r-lib/credentials/pull/7. I'm not actually proposing this function name. I still think it's possible that this function belongs in credentials? In any case, that needs to be discussed in order to pick a name and decide where/if to export.

I have confirmed that things "just work" for me with GitHub.com and GitHub.ubc.ca.


DESCRIPTION
```diff
@@ -20,16 +20,23 @@ URL: https://github.com/r-lib/gh#readme
 BugReports: https://github.com/r-lib/gh/issues
 Imports:
     cli (>= 2.0.1),
+    curl,
```
jennybc, line 23: Temporary. If `set_github_pat2()` stays here, we can do token validation without a direct use of curl, i.e. using only gh functions.

DESCRIPTION
```diff
@@ -20,16 +20,23 @@ URL: https://github.com/r-lib/gh#readme
...
     httr (>= 1.2),
     ini,
-    jsonlite
+    jsonlite,
+    stats,
+    withr
```
jennybc, line 28: New Import. If `set_github_pat2()` goes to credentials, can move back to suggests.

R/gh_request.R
```diff
@@ -108,7 +108,7 @@ gh_set_url <- function(x) {
     x$url <- URLencode(x$endpoint)
     x$api_url <- get_baseurl(x$url)
   } else {
-    x$api_url <- x$api_url %||% default_api_url()
+    x$api_url <- get_apiurl(x$api_url %||% default_api_url())
```
jennybc, line 111: Makes it so we can (quietly) accept "https://github.com" or "https://github.acme.com" instead of insisting on "https://api.github.com" or "https://github.acme.com/api/v3".

R/gh_token.R
```diff
@@ -83,7 +83,21 @@
 gh_token <- function(api_url = NULL) {
   api_url <- api_url %||% default_api_url()
   token_env_var <- paste0("GITHUB_PAT_", slugify_url(api_url))
-  get_first_token_found(c(token_env_var, "GITHUB_PAT", "GITHUB_TOKEN"))
+  candidates <- c(
+    token_env_var,
+    if (is_github_dot_com(api_url)) c("GITHUB_PAT", "GITHUB_TOKEN")
```
jennybc, line 88: This makes it so we no longer send a GITHUB_PAT to a server other than GitHub.com. I saw this in credentials/gert and I think it's a good idea here too.

R/gh_token.R
```diff
@@ -83,7 +83,21 @@
...
+  )
+  val <- get_first_token_found(candidates)
+  if (val != "" || !can_load("credentials")) {
+    return(val)
+  }
+
+  set_github_pat2(api_url)
```
jennybc, line 95: The main new function, which is adapted from `credentials::set_github_pat()`, hence it's temporary name.

R/gh_token.R
```diff
@@ -135,8 +151,8 @@ get_first_token_found <- function(vars) {
     if (has_keyring) tryCatch(keyring::key_get(v), error = function(e) NULL)
   }
   for (var in vars) {
-    if ((val <- key_get(var) %||% "") != "") break
     if ((val <- Sys.getenv(var, "")) != "") break
+    if ((val <- key_get(var) %||% "") != "") break
```
jennybc, line 155: This closes #110, since I'm here.

R/gh_token.R
```diff
@@ -164,21 +180,162 @@ gh_auth <- function(token) {
...
+        message2(sprintf(
+          "To generate a new token, visit: %s/settings/tokens",
+          host_url
+        ))
+        credentials::credential_reject(cred)
+        next
+      }
+      if (isTRUE(validate)) {
+        hx <- curl::handle_setheaders(curl::new_handle(), Authorization = paste("token", cred$password))
+        req <- curl::curl_fetch_memory(sprintf("%s/user", api_url), handle = hx)
```
jennybc, line 305: This can be done with gh technology if this stays here vs. using curl directly.

R/gh_token.R
```diff
@@ -164,21 +180,162 @@ gh_auth <- function(token) {
...
 }
 
-get_baseurl <- function(x) {
-  if (!any(grepl("^https?://", x))) stop("Only works with HTTP(S) protocols")
-  prot <- sub("^(https?://).*$", "\\1", x)
-  rest <- sub("^https?://(.*)$", "\\1", x)
-  host <- sub("/.*$", "", rest)
-  paste0(prot, host)
+# taken directly from credentials
+message2 <- function(...){
```
jennybc, line 338: I have not thought much about why credentials overrides `message()` (?). I just copied it over.

R/gh_token.R
```diff
@@ -164,21 +180,162 @@ gh_auth <- function(token) {
...
+        ),
+        GIT_ASKTOKEN_NAME = sprintf(
+          "Personal Access Token (PAT) for %s", host_url
+        )
+      ))
+    }
+  } else {
+    withr::local_envvar(c(
+      GIT_ASKPASS = system.file(
+        "empty_token.sh", package = "gh", mustWork = TRUE
```
jennybc, line 286: This is the only, clean way I have found to reliably prevent an attempt to interact with the user in semi-interactive settings, i.e. when testing or checking.

R/gh_request.R
jennybc: This makes sense as a global attitude, since GitHub.com PATs are already keyed with "https://github.com" in the Git credential store, as opposed to "https://api.github.com". I think we should extend the same policy to GHE PATs and, since we do token lookup based on `.api_url`, we eventually arrive at this.

R/gh_token.R
```diff
@@ -164,21 +180,162 @@ gh_auth <- function(token) {
 #' # A custom one
 #' slugify_url("https://github.acme.com")
 
-slugify_url <- function(url) {
+slugify_url <- function(url) {           # https://jane@github.uni.edu/api/v3
+  url <- get_baseurl(url)                # https://jane@github.uni.edu
```
jennybc, line 184: @gaborcsardi this is a slightly new behaviour. It means the slug is GITHUB_UNI_EDU now, instead of GITHUB_UNI_EDU_API_V3.

R/gh_token.R
```diff
@@ -164,21 +180,162 @@ gh_auth <- function(token) {
...
+  url <- normalize_host(url)
+  x2 <- sub("^.*://([^/]*@)?", "", url)  # github.uni.edu
+  x3 <- gsub("[.]+", "_", x2)            # github_uni_edu
```
jennybc, line 187: @gaborcsardi This regex changed (lost the `/`) because we've already truncated to the base url and should no longer have any `/`.

R/gh_token.R
```diff
@@ -164,21 +180,162 @@ gh_auth <- function(token) {
...
+normalize_host <- function(x) {
+  sub("api.github.com", "github.com", x)
+}
+
+get_hosturl <- function(url) {
+  url <- get_baseurl(url)
+  normalize_host(url)
+}
+
+get_apiurl <- function(url) {
```
jennybc, line 211: Since we accept "https://github.com" and "https://github.ubc.ca" instead of insisting on "https://api.github.com" and "https://github.ubc.ca/api/v3", we need this function to ensure we have the true API root url.

DESCRIPTION
gaborcsardi: we can use curl, no problem.

DESCRIPTION
gaborcsardi: withr is fine as well

R/gh_token.R
```diff
@@ -152,33 +168,182 @@ gh_auth <- function(token) {
...
+    return(FALSE)
+  }
+
+  host_url <- get_hosturl(api_url)
+  prot_host <- strsplit(host_url, "://", fixed = TRUE)[[1]]
+  if (is_github_dot_com(api_url)) {
+    # what's currently in credentials
+    # pat_user <- Sys.getenv("GITHUB_PAT_USER", 'PersonalAccessToken')
+    # TODO: I propose we change the user env var to GITHUB_USER
+    user_env_var <- "GITHUB_PAT_USER"
```
gaborcsardi, line 252: Why do we actually need the user? If we use a PAT, then no user is needed, right? If we get the auth info from the credential store, that'll tell us the user?

R/gh_token.R
gaborcsardi: I don't think this'll work on Windows. But we can think about a different solution later.

R/gh_token.R
gaborcsardi: I am not sure if we need to validate the token. The real query might fail, anyway, the token can be invalidated externally, etc. so errors must be handled there.

R/gh_token.R
gaborcsardi: This is probably because sys/git writes to the console directly, and if we don't flush then the R output and the sys/git output is mixed up.

R/gh_token.R
```diff
@@ -152,33 +168,182 @@ gh_auth <- function(token) {
...
+    if(isTRUE(verbose)) {
+      message2("If prompted for GitHub credentials, enter your PAT in the password field")
+    }
+    askpass <- Sys.getenv('GIT_ASKPASS')
+    if (nchar(askpass)) {
+      # Hack to override prompt sentence to say "Token" instead of "Password"
+      withr::local_envvar(c(
+        GIT_ASKTOKEN = askpass,
+        GIT_ASKPASS = system.file(
+          "ask_token.sh", package = "gh", mustWork = TRUE
```
gaborcsardi, line 284: I don't think this will work on Windows in general. It probably works for CRAN, because they have `sh`, but many people don't.

R/gh_token.R
jennybc: I am just copying what `credentials::set_github_pat()` does. I can provide some context, though.

This is a faithful implementation of the official workflow recommended in "Typical Use of Git Credential":

https://git-scm.com/docs/git-credential#_typical_use_of_git_credential

It makes sense to validate right after the `git credential fill` call, which happens inside `credentials::git_credential_ask(pat_url, verbose = verbose)` for us.  That fetches a matching credential from the store or prompts the user for the PAT. Then you immediately do either `git credential approve` (valid PAT)  or `git credential reject` (invalid PAT). Implications:

* New PAT we just from user: By validating, we make sure we don't put a bad PAT into the store.
* Existing PAT we got from the user If PAT is good, it persists, but if it's known to have gone bad, we remove it.

This seems worthwhile to me.

Yeah, it's also true that a subsequent request can fail, but remember these are all happening extremely close in time. This validation only happens once per session, when we move the PAT from the store to an env var. Token invalidation between pulling from the store and usage seems unlikely. More likely is the token having the wrong scopes vis-a-vis a particular request.

R/gh_token.R
jennybc: Maybe @jeroen can tell us. A very similar method is used in the other branch (interactive session), so I don't see how that could work and this could not?

gaborcsardi: Neither work. You can't rely on `sh` on Windows.

R/gh_token.R
jennybc: See this part of the vignette:

https://github.com/r-lib/gh/blob/9d188d3739af6adf77d70b8324260a2b5d71ed2a/vignettes/Managing-Personal-Access-Tokens.Rmd#L171-L189

@jeroen has convinced me this is correct. In my previous PR (https://github.com/r-lib/credentials/pull/7), I also wanted to set it to the actual username. But then it will clobber a username/password user may have cached through normal command line git via https.

Using 'PersonalAccessToken' is what [Git Credential Manager Core](https://github.blog/2020-07-02-git-credential-manager-core-building-a-universal-authentication-experience/) does.

The `user` also provides a non-obvious and not-documented way to use more than 1 PAT for a given GitHub host.

jennybc:
A really big question to me is:

Where should this function live? credentials or gh?

If credentials, does it replace / generalize `credentials::set_github_pat()`?

If gh, what do we call it?

R/gh_token.R
gaborcsardi: Yes, what I meant is that we do not need the env var at all, and we don't need to worry about the username at all.

The vignettes discusses to use the username for multiple credentials, but I don't think we need to worry about this, other then possible allowing a username in the URL.

gaborcsardi:
> Where should this function live? credentials or gh?

They are fine here, I think. Jeroen can always add them to credentials later if he wants.

R/gh_token.R
jeroen: The git for windows credential for Windows has its own GUI. It completely ignores these environment variables.

jeroen: Also, note that Git for Windows brings its own bash shell environment and everything, because git depends on this. I don't think it's possible to have a working git installation, but not sh.

gaborcsardi: Sure, but `git` is on the path, and `sh` is not.

gaborcsardi: But let's merge it and we'll see if it works. :)

jennybc:
I'm going to close this, now that the URL aspects seem likely to appear in credentials.

I'll be back with yet another PR after that to implement the bits that still remain in gh.

PR #129, opened 2020-08-08
Introduce gh_pat class

### LINKED ISSUE #125: Use an S3 class and print method for PATs
I propose we introduce an S3 class for PATs, just so we can give them a coy print method, i.e. not reveal the whole PAT.

I think this is probably "worth it" in terms of helping users and maybe even us to not leak a PAT via, e.g., rendered documentation or a log file.

jennybc:
Closes #125

What do you think the name of this class should be? It really doesn't matter. I contemplated `gh_token` but then I can't make `gh_token()` the constructor, because `gh_token()` already exists as a user-facing, higher-level function.


jennybc:
Here's how it looks with no token:

``` r
library(gh)

gh_token()
#> <no PAT>

format(gh_token())
#> [1] "<no PAT>"

str(gh_token())
#> <gh_pat> <no PAT>
```

<sup>Created on 2020-08-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

And with:

``` r
library(gh)

credentials::set_github_pat()

gh_token()
#> e5fa...44

format(gh_token())
#> [1] "e5fa...44"

str(gh_token())
#> <gh_pat> e5fa...44
```

<sup>Created on 2020-08-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

PR #130, opened 2020-08-13
Formalize the GitHub PAT search strategy

jennybc:
Laying the groundwork to incorporate a URL-based query to the Git credential store.


jennybc:
Major issues:

* If "git" fails, I really think this should return `""` instead of failing. We need to find a way to make `credentials::set_github_pat()` (or any equivalent that we write) act more like a library and return an error code vs. stopping execution itself.
* The "git" strategy doesn't work for deployments other than "GitHub.com" yet.
* `strategy` probably needs to be controllable via an env var. Maybe `GH_PAT_STRATEGY`?

PR #132, opened 2020-09-13
Gitcreds

gaborcsardi:


R/gh_token.R
```diff
@@ -7,20 +7,23 @@
...
-#' [gitcreds::gitcreds_get()] function to see if this is the case. This
-#' function errors if it cannot find any PAT.
+#' gh calls [gitcreds::gitcreds_get()] with the `api_url`, which checks session
+#' environment variables and then the local Git credential store for a PAT
+#' appropriate to the `api_url`. Therefore, if you have previously used a PAT
+#' with, e.g., command line Git, gh may retrieve and re-use it. You can call
+#' [gitcreds::gitcreds_get()] directly, yourself, if you want to see what is
+#' found for a specific URL. If no matching PAT is found,
+#' [gitcreds::gitcreds_get()] errors, whereas `gh_token()` does not and,
+#' instead, returns `""`.
```
jennybc, line 17: I rephrased this somewhat. Because if you've used command line Git with USERNAME/PASSWORD, gitcreds should **not** be finding those creds. Only if the creds went it with some PAT-y workflow. The only people for whom these are the same are people using 2FA.

Or, if I'm wrong, then I predict some trouble with a password getting sent as a PAT for people who aren't using 2FA and their password != a PAT.

jennybc (approved):

R/gh_token.R
gaborcsardi: Yeah, gitcreds returns any credentials, token or not. It is up to gh and other upstream packages to decide what to do with the credentials.

gaborcsardi: Btw. how is 2FA related to this?

gaborcsardi:
@jennybc Thanks for the fixes!

R/gh_token.R
jennybc: If you've turned on 2FA for GitHub, any https access MUST occur with a PAT, i.e. you cannot ever use username/password.

https://docs.github.com/en/github/authenticating-to-github/accessing-github-using-two-factor-authentication#using-two-factor-authentication-with-the-command-line

jennybc: And other people (who may not have turned on 2FA) _may_ behave similarly. So this is a key part of why https is encouraged by gert and usethis. Because once we can get as much agreement as possible between gert and gh and between (gert+gh) and command line git, a huge chunk of users can set up a PAT and several tools all start to "just work".

Issue #135, opened 2020-10-17
Weird behavior with .limit on actions api endpoints


rundel:
I've run into some weird behavior when trying to set a limit for results coming from the actions api (it may affect other areas - I haven't tested beyond these endpoints). Seems to be a problem with the paging, but I haven't dug into the underlying code.

```
ok = gh::gh(
  "GET /repos/:owner/:repo/actions/runs",
  owner = "r-lib",
  repo = "gh"
)

error = gh::gh(
  "GET /repos/:owner/:repo/actions/runs",
  owner = "r-lib",
  repo = "gh",
  .limit = 1
)

badly_formed = gh::gh(
  "GET /repos/:owner/:repo/actions/runs",
  owner = "r-lib",
  repo = "gh",
  .limit = 10
)
```

gaborcsardi:
So, #136 fixes the `badly_formed` query, but the `error` one is still an error, is that expected?

rundel:
I don't think that is expected - I clearly didn't test against my own test cases.

rundel:
I think the issue is around comparing `length(res)` to `.limit` since the former is not correct for the "named result" case

gaborcsardi:
Yeah, we should probably measure the length differently in the "named result" case.

gaborcsardi:
OTOH, this is in some sense a different issue from the one you fixed.

rundel:
I think the while loop bit can be fixed by counting pages and then checking that `pages * per_page < .limit`, the harder bit is with the subsetting to obtain the `.limit` length in the final conditional because it is not obvious which of the elements contains the results.

gaborcsardi:
> I think the while loop bit can be fixed by counting pages and then checking that pages * per_page < .limit

Yeah, although th API is not guaranteed to give you `per_page` results. Still better than the current behavior I guess.

> not obvious which of the elements contains the results.

Yeah.

gaborcsardi:
@rundel Can we use `total_count` to inform us about the response type? 

Btw. there is an API spec now, which will make our lives easier, even though it is massive: https://github.com/github/rest-api-description/tree/main/descriptions/api.github.com

rundel:
I wasn't sure if `total_count` was consistent across the api - I feel like there were a few cases that used something else, but I may be confusing that the with result unnamed array.

rundel:
The API spec looks like it might be useful - it seems like within the schema definition they distinguish between "object" results vs "array" results.

[...]

PR #136, opened 2020-10-29
Pagination merge issue

rundel:
This is related to #135, I was also seeing strange behavior with the search API and I think I understand the cause now.

Most of the GitHub api endpoints either return a single result as a single json object (e.g. `GET /user`) or multiple results as a json array (e.g. `GET /orgs/:org/members`). However, there are a few endpoints where the endpoint returns a single json object that then contains the paginated array (e.g. `GET /repos/:owner/:repo/actions/runs` and `GET /search/repositories`) - the problem arises for these latter cases.

This pull request has a proposed fix to deal with the merging of these types separately so that we get the "right" behavior for all three cases. The bug is a bit insidious because any results past the first page are concatenated to the result object, but because the attributes are overwritten the new values don't have a `names` attribute and the first results are present with the expected structure. See the example below:

```r
z = gh::gh(endpoint = "GET /search/repositories", q = "org:r-lib", .limit = 200)

str(z, max.level=1, give.attr = FALSE)
## List of 6
##  $ total_count       : int 115
##  $ incomplete_results: logi FALSE
##  $ items             :List of 100
##  $ NA                : int 115
##  $ NA                : logi FALSE
##  $ NA                :List of 15
```

Additionally, I think some rethinking of the pagination may be necessary. For example `length(res) < .limit` in https://github.com/r-lib/gh/blob/b524a442b854e9acc45fed0b0f3e5b2666fc8783/R/gh.R#L177 does not work correctly for  the named multiresult endpoints but is also doing the wrong thing for the single result endpoints (although this doesn't matter since these never have a `next` link). The name of the results array also changes between endpoints so it is difficult to workout what the current # of results is.

Another related problem is using a small `.limit` with one of the single result endpoints, e.g.
```r
gh::gh(endpoint = "GET /user", .limit = 1)
Error in attributes(res) <- res_attr : 
  'names' attribute [39] must be the same length as the vector [1]
```
as this currently ends up incorrectly subsetting the results object values instead of the "results".

### DIFF
--- a/R/gh.R
+++ b/R/gh.R
@@ -177,7 +177,20 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
   while (!is.null(.limit) && length(res) < .limit && gh_has_next(res)) {
     if (.progress) update_progress_bar(prbr, res)
     res2 <- gh_next(res)
-    res3 <- c(res, res2)
+
+    if (!is.null(names(res2)) && identical(names(res), names(res2))) {
+      res3 <- mapply(           # Handle named array case
+        function(x, y) {        # e.g. GET /search/repositories
+          z <- c(x, y)
+          if (is.atomic(z)) unique(z)
+          else z
+        },
+        res, res2
+      )
+    } else {                    # Handle unnamed array case
+      res3 <- c(res, res2)      # e.g. GET /orgs/:org/invitations
+    }
+
     attributes(res3) <- attributes(res2)
     res <- res3
   }
--- a/R/gh.R
+++ b/R/gh.R
@@ -185,7 +185,8 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
           if (is.atomic(z)) unique(z)
           else z
         },
-        res, res2
+        res, res2,
+        SIMPLIFY = FALSE
       )
     } else {                    # Handle unnamed array case
       res3 <- c(res, res2)      # e.g. GET /orgs/:org/invitations


R/gh.R
```diff
@@ -177,7 +177,20 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
   while (!is.null(.limit) && length(res) < .limit && gh_has_next(res)) {
     if (.progress) update_progress_bar(prbr, res)
     res2 <- gh_next(res)
-    res3 <- c(res, res2)
+
+    if (!is.null(names(res2)) && identical(names(res), names(res2))) {
+      res3 <- mapply(           # Handle named array case
+        function(x, y) {        # e.g. GET /search/repositories
+          z <- c(x, y)
+          if (is.atomic(z)) unique(z)
```
rundel, line 185: Not sure this is necessary, but it seemed the cleanest option based on the relevant endpoints

gaborcsardi: When do we actually need this?

R/gh.R
```diff
@@ -177,7 +177,20 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
   while (!is.null(.limit) && length(res) < .limit && gh_has_next(res)) {
     if (.progress) update_progress_bar(prbr, res)
     res2 <- gh_next(res)
-    res3 <- c(res, res2)
+
+    if (!is.null(names(res2)) && identical(names(res), names(res2))) {
+      res3 <- mapply(           # Handle named array case
```
gaborcsardi, line 182: Do we need `SIMPLIFY = FALSE` for `mapply()`?

gaborcsardi (commented):
Thanks!

R/gh.R
rundel: The results from the search api look the following:
```
{
  "total_count": 40,
  "incomplete_results": false,
  "items": [
    {
      ...
    }
  ]
}
```

when there are multiple pages the `items` will change but `total_count` and `incomplete_results` are the same for all of the pages, hence collapsing them with unique. It seems like for this case of response the non-array values are meant to give overall context, but this is purely supposition on my part.

R/gh.R
rundel: That seems like a good idea

gaborcsardi:
Thanks! This sort of heuristic is somewhat risky, because we don't know what it might break in the other parts of the API. OTIH it seems beneficial, so I'll merge it and we'll if it breaks anything.

PR #138, opened 2020-11-09
Add function to authenticate as GitHub App

jeroen:
This makes it possible to create GitHub Apps in R. 

The new dependency `jose` is an R-only package (no compiled code), which depends only on jsonlite and openssl, which we already depend on via httr. So it is very light.


gaborcsardi:
Cool, thanks. Can you please use gh's functions to build and perform the request, etc.? Thanks!

jeroen:
One thing that we could consider, is to integrate this further into gitcreds. Such that when a `GH_APP_ID` and `GH_APP_KEY` environment variables are set, we automatically generate a token.

That way people can switch between setting a `GITHUB_PAT` or `GH_APP_ID`+`GH_APP_KEY` in their CI without making any code changes in their `gh() `calls.

jeroen:
Given that we don't need any internal gh functions for this, it may as well live in a separate package.

PR #141, opened 2020-11-27
Fix pagination for various response types

### LINKED ISSUE #135: Weird behavior with .limit on actions api endpoints

gaborcsardi:
Use 'total_count' and info from the schema on
how to calculate the "length" of a response.

Fixes #135.

### DIFF
--- a/R/gh.R
+++ b/R/gh.R
@@ -176,8 +176,9 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
   raw <- gh_make_request(req)
 
   res <- gh_process_response(raw)
+  len <- gh_response_length(res)
 
-  while (!is.null(.limit) && length(res) < .limit && gh_has_next(res)) {
+  while (!is.null(.limit) && len < .limit && gh_has_next(res)) {
     if (.progress) update_progress_bar(prbr, res)
     res2 <- gh_next(res)
 
@@ -195,17 +196,37 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
       res3 <- c(res, res2)      # e.g. GET /orgs/:org/invitations
     }
 
+    len <- len + gh_response_length(res2)
+
     attributes(res3) <- attributes(res2)
     res <- res3
   }
 
-  if (! is.null(.limit) && length(res) > .limit) {
+  if (! is.null(.limit) && len > .limit) {
     res_attr <- attributes(res)
     res <- res[seq_len(.limit)]
     attributes(res) <- res_attr
   }
 
   res
+}
+
+gh_response_length <- function(res) {
+  if (!is.null(names(res)) && length(res) > 1 &&
+      names(res)[1] == "total_count") {
+    # Ignore total_count, incomplete_results, repository_selection
+    # and take the first list element to get the length
+    lst <- vapply(res, is.list, logical(1))
+    nm <- setdiff(
+      names(res),
+      c("total_count", "incomplete_results", "repository_selection")
+    )
+    tgt <- which(lst[nm])[1]
+    if (is.na(tgt)) length(res) else length(res[[ nm[tgt] ]])
+    length(res[[2]])
+  } else {
+    length(res)
+  }
 }
 
 gh_make_request <- function(x) {
--- a/R/gh.R
+++ b/R/gh.R
@@ -208,7 +208,9 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
     res <- res3
   }
 
-  if (! is.null(.limit) && len > .limit) {
+  # We only subset for a non-named response.
+  if (! is.null(.limit) && len > .limit &&
+      ! "total_count" %in% names(res) && length(res) == len) {
     res_attr <- attributes(res)
     res <- res[seq_len(.limit)]
     attributes(res) <- res_attr
--- a/R/gh.R
+++ b/R/gh.R
@@ -223,7 +223,6 @@ gh_response_length <- function(res) {
     )
     tgt <- which(lst[nm])[1]
     if (is.na(tgt)) length(res) else length(res[[ nm[tgt] ]])
-    length(res[[2]])
   } else {
     length(res)
   }
--- a/R/gh.R
+++ b/R/gh.R
@@ -184,12 +184,18 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
 
     if (!is.null(names(res2)) && identical(names(res), names(res2))) {
       res3 <- mapply(           # Handle named array case
-        function(x, y) {        # e.g. GET /search/repositories
+        function(x, y, n) {        # e.g. GET /search/repositories
           z <- c(x, y)
-          if (is.atomic(z)) unique(z)
-          else z
+          atm <- is.atomic(z)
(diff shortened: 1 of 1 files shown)


R/gh.R
```diff
@@ -195,11 +196,13 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
...
+    len <- len + gh_response_length(res2)
+
     attributes(res3) <- attributes(res2)
     res <- res3
   }
 
-  if (! is.null(.limit) && length(res) > .limit) {
+  if (! is.null(.limit) && len > .limit) {
     res_attr <- attributes(res)
     res <- res[seq_len(.limit)]
```
rundel, line 207: There needs to be alternative behavior here since for the object case we need to subset the array within the object

R/gh.R
```diff
@@ -208,6 +211,24 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
...
+    # and take the first list element to get the length
+    lst <- vapply(res, is.list, logical(1))
+    nm <- setdiff(
+      names(res),
+      c("total_count", "incomplete_results", "repository_selection")
+    )
+    tgt <- which(lst[nm])[1]
+    if (is.na(tgt)) length(res) else length(res[[ nm[tgt] ]])
+    length(res[[2]])
+  } else {
```
rundel, line 226: I think there needs to be a third case here for endpoints like `/user`

R/gh.R
```diff
@@ -208,6 +211,24 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
...
 
+gh_response_length <- function(res) {
+  if (!is.null(names(res)) && length(res) > 1 &&
+      names(res)[1] == "total_count") {
+    # Ignore total_count, incomplete_results, repository_selection
+    # and take the first list element to get the length
+    lst <- vapply(res, is.list, logical(1))
+    nm <- setdiff(
+      names(res),
+      c("total_count", "incomplete_results", "repository_selection")
```
rundel, line 222: I assume these are coming from the API description?

R/gh.R
```diff
@@ -208,6 +211,24 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
...
+    )
+    tgt <- which(lst[nm])[1]
+    if (is.na(tgt)) length(res) else length(res[[ nm[tgt] ]])
+    length(res[[2]])
```
rundel: Logic above seems a bit wonky? Seems like `length(res[[2]])` or the conditional should be there not both.

R/gh.R
```diff
@@ -208,6 +211,24 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
   res
 }
 
+gh_response_length <- function(res) {
+  if (!is.null(names(res)) && length(res) > 1 &&
+      names(res)[1] == "total_count") {
```
rundel, line 216: Is `total_count` always the first entry or would it be better to just use `%in%`?

R/gh.R
```diff
@@ -208,6 +211,24 @@ gh <- function(endpoint, ..., per_page = NULL, .token = NULL, .destfile = NULL,
...
+    # Ignore total_count, incomplete_results, repository_selection
+    # and take the first list element to get the length
+    lst <- vapply(res, is.list, logical(1))
+    nm <- setdiff(
+      names(res),
+      c("total_count", "incomplete_results", "repository_selection")
+    )
+    tgt <- which(lst[nm])[1]
```
rundel, line 224: Include some kind of runtime assert that checks that there is only one target? If not throw an error and ask the user to report the endpoint?

R/gh.R
gaborcsardi: This was a typo, removed the `length(res[[2]])` line.

R/gh.R
gaborcsardi: It seems like it is the first entry, yeah.

R/gh.R
gaborcsardi: Yeah, but it is not entirely clear how to do that. This branch is only used if we got more items than requested, which might not even happen. So this is not super important I think. I'll protect this with `length(res) == len`, so we only do it in the non-named response case.

rundel: I was thinking of a case where someone set `.limit = 101` in which case they would now get 200 results (not the end of the world).

Can we not use the same logic to find the "target" as the unnamed list and subset it?

R/gh.R
gaborcsardi: The length is only really important if `.limit` is specified, and people will not do that for `/user`. 

We could do more heuristics, but I want to be cautious. So for `/user` we can essentially keep the old behavior, nothing changes imo.

R/gh.R
gaborcsardi: yeah

R/gh.R
gaborcsardi: Well, you PR merges all elements, and I looked over the spec, so this should be fine, I think. If there are multiple of them, we'd still want to do something, so I think taking the first is fine.

rundel: Makes sense, particularly if we're just looking for the length

R/gh.R
gaborcsardi: Yeah, OTOH subsetting based on some target list is potentially destructive, while just looking up the length os probably not. 

So I think we could only subset in the non-named case, to keep the old behavior for that. The behavior for the named case is much improved already, so it is fine I think.

R/gh.R
rundel: This seems reasonable, I will need to check that ghclass is not doing something silly like setting `.limit` for everything

R/gh.R
rundel: Sounds good - I think the usage of `.limit` is quite rare already and this should cover the vast majority of use cases

gaborcsardi:
@rundel Thanks for looking at this. I think it is definitely an improvement and hopefully we were conservative enough not to break anyone's code.

There is still this to fix:
```r
se = gh::gh(
  "GET /search/code?q=foobar",
  .limit=101
)
lengths(se)
#>        total_count incomplete_results              items
#>                  2                  2                200

se$total_count
#> [1] 18043087 18080930

se$incomplete_results
#> [1]  TRUE FALSE
```

In the future I am hoping that we can use r-lib/webfakes and the info from the spec to write a test suite with reasonable coverage, and then we can be more confident about handling the output of the API better.

gaborcsardi:
Oh, yeah the issue with that code is this line:
```
          if (is.atomic(z)) unique(z)
```
and of course the API provided different values for `total_count` and `incomplete_results` for the two responses. I'll tweak that a bit and will use the latest value for these two fields.

PR #144, opened 2020-12-02
Warn if a token contains whitespace characters

### LINKED ISSUE #143: Basic token validation
Somewhat related to https://github.com/r-lib/gh/issues/133 but simpler:

I have been debugging a `GitHub API error (422): 422 Unprocessable Entity` for a long time. The problem was that the GHA UI where you have to save secrets had appended a `\n` at the end of the secret. Apparently this makes the token invalid.

It would make sense to have `gh` do some basic validation on the token format, before calling to the API. Something as simple as `grepl('[a-f0-9]{40}$',  token)` could work.

As a side node, gh-apps tokens are prefixed with `v1.` and then the usual hash.

jeroen:
Fixes #143

### DIFF
--- a/R/gh_token.R
+++ b/R/gh_token.R
@@ -54,7 +54,10 @@ gh_token <- function(api_url = NULL) {
 
 gh_auth <- function(token) {
   if (isTRUE(token != "")) {
-    c("Authorization" = paste("token", token))
+    if(any(grepl("\\W", token))) {
+      warning("Token contains whitespace characters")
+    }
+    c("Authorization" = paste("token", trim_ws(token)))
   } else {
     character()
   }


gaborcsardi:
Thanks!

jennybc:
In one of the PRs I worked on (now I can't remember which repo ... credentials? gh? gitcreds?) that didn't get merged, there were quite a few tests of the PAT. I.e. a certain number of hexadecimal characters. Are we using those anywhere?

jeroen:
In credentials we do some extra tests in `set_github_pat()` specifically for personal access tokens. But as Gabor notes in https://github.com/r-lib/gh/issues/143, different types of tokens have different formats, which may change in the future, so we cannot be overly strict.

jeroen:
Update: actually it seems gh indeed has some validation already, but I am not sure where exactly this gets applied. 

https://github.com/r-lib/gh/blob/978cb6e7e934a384aac2c63a090cc0b9ffb6e057/R/gh_token.R#L67-L87

gaborcsardi:
Maybe we just need a `^` and `$` in the regexp.

jeroen:
But in that case it is overly strict, because some tokens are prefixed with `v1.`

PR #149, opened 2021-04-01
update PAT validation for new token formats

### LINKED ISSUE #148: Update PAT validation
https://github.blog/changelog/2021-03-04-authentication-token-format-updates/

PATs are included in in "each of these token types" here:

> For each of these token types we are making the following changes:
> 
>   * The character set is changing from [a-f0-9] to [A-Za-z0-9_]
>   * The format is changing to include a prefix:
>     - gp1_ for Personal Access Tokens
>   ...
>
>   The overall length of our tokens will remain the same for now. However, GitHub tokens will likely increase in length in future updates, so integrators should plan to support tokens up to 255 characters after June 1, 2021.

This needs to change:

https://github.com/r-lib/gh/blob/978cb6e7e934a384aac2c63a090cc0b9ffb6e057/R/gh_token.R#L75-L83

fmichonneau:
fix #148 

update the regular expression to match the new token formats.

* Do we want to allow for the non-PAT prefixes? in other words, should only `ghp_` prefixes be allowed? (this PR allows the 5 prefixes mentioned in the [blog post](https://github.blog/changelog/2021-03-04-authentication-token-format-updates/)
* Should [tests](https://github.com/r-lib/gh/blob/master/tests/testthat/test-gh_token.R#L59-L63) be modified to account for this regular expression change?
* Do we want to already allow for up to 255 characters?

### DIFF
--- a/R/gh_token.R
+++ b/R/gh_token.R
@@ -78,7 +78,7 @@ validate_gh_pat <- function(x) {
   if (x == "" || grepl("^gh[pousr]_[A-Za-z0-9_]{36}$", x) || grepl("[[:xdigit:]]{40}", x)) {
     x
   } else {
-    throw(new_error("A GitHub PAT must consist of 40 hexadecimal digits"))
+    throw(new_error("A GitHub PAT must consist of 40 characters."))
   }
 }
 
--- a/R/gh_token.R
+++ b/R/gh_token.R
@@ -79,17 +79,25 @@ new_gh_pat <- function(x) {
   if (is.character(x) && length(x) == 1) {
     structure(x, class = "gh_pat")
   } else {
-    throw(new_error("A GitHub PAT must be a string"))
+    throw(new_error("A GitHub PAT must be a string", call. = FALSE))
   }
 }
 
 # validates PAT only in a very narrow, technical, and local sense
 validate_gh_pat <- function(x) {
   stopifnot(inherits(x, "gh_pat"))
-  if (x == "" || grepl("^gh[pousr]_[A-Za-z0-9_]{36,251}$", x) || grepl("[[:xdigit:]]{40}", x)) {
+  if (x == "" ||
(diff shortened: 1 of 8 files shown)


jennybc:
* I think it makes sense to accept the non-PAT prefixes
* I would accept up to 255 characters in the regular expression that has the prefixes. Intent: I hope this means "tokens created under the new regime".
* I would leave the exact length requirement of 40 for the regular expression targeting "legacy" PATs. This has been very useful for catching people providing a regular password as PAT.
* Seems like we should add a test with a new-style PAT to make sure it does NOT error.

R/gh_token.R
```diff
@@ -75,10 +75,10 @@ new_gh_pat <- function(x) {
 # validates PAT only in a very narrow, technical, and local sense
 validate_gh_pat <- function(x) {
   stopifnot(inherits(x, "gh_pat"))
-  if (x == "" || grepl("[[:xdigit:]]{40}", x)) {
+  if (x == "" || grepl("^gh[pousr]_[A-Za-z0-9_]{36}$", x) || grepl("[[:xdigit:]]{40}", x)) {
```
jennybc, line 78:
```suggestion
  if (x == "" || grepl("^gh[pousr]_[A-Za-z0-9_]{36,251}$", x) || grepl("[[:xdigit:]]{40}", x)) {
```

R/gh_token.R
```diff
@@ -75,10 +75,10 @@ new_gh_pat <- function(x) {
...
     x
   } else {
-    throw(new_error("A GitHub PAT must consist of 40 hexadecimal digits"))
+    throw(new_error("A GitHub PAT must consist of 40 characters."))
```
jennybc, line 81: This wording will have to get even more general.

fmichonneau:
let me know what you think of these new changes. Thanks!

jennybc:
Thanks!

fmichonneau:
Thank you!

PR #151, opened 2021-04-01
Update with current errors.R from processx

jennybc:
I did this because it feels like `call.` does not have the usual effect. I figured this has since been fixed in the compat file. But even after I update, `call.` doesn't work the way I expect.

Is this expected behaviour? I'm talking about the `Error in FALSE` part.

``` r
gh:::gh_pat("aa")
#> Error in FALSE: GitHub PAT must have one of these forms:
#>   * 40 hexadecimal digits (older PATs)
#>   * A 'ghp_' prefix followed by 36 to 251 more characters (newer PATs)
```

<sup>Created on 2021-04-01 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0.9002)</sup>

Generated by https://github.com/r-lib/gh/blob/86eb3d975bef9133b62c82176c2d893ddb4110da/R/gh_token.R#L84-L89

### DIFF
--- a/R/errors.R
+++ b/R/errors.R
@@ -89,6 +89,10 @@
 # ### 1.2.3 -- 2021-03-06
 #
 # * Use cli instead of crayon
+#
+# ### 1.2.4 -- 2012-04-01
+#
+# * Allow omitting the call with call. = FALSE in `new_cond()`, etc.
 
 err <- local({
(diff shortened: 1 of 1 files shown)


gaborcsardi:
Thanks!

PR #157, opened 2021-08-13
New example of graphql query

statnmap:
Proposition of use for graphql query to retrieve the graphql rate limit, different from the REST API one as explained here: https://docs.github.com/en/graphql/overview/resource-limitations

_I will let you recreate the documentation as it needs {roxygenlabs} that I dont have._

### DIFF
--- a/R/gh_gql.R
+++ b/R/gh_gql.R
@@ -26,7 +26,7 @@
 #'   }
 #' }"
 #' 
-#' gh::gh_gql(ratelimit_query)
+#' gh_gql(ratelimit_query)
 
 gh_gql <- function(query, ...) {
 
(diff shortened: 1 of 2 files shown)


R/gh_gql.R
```diff
@@ -13,6 +13,20 @@
...
+#'   }
+#'   rateLimit {
+#'     limit
+#'     cost
+#'     remaining
+#'     resetAt
+#'   }
+#' }"
+#' 
+#' gh::gh_gql(ratelimit_query)
```
gaborcsardi, line 29: Thanks! Can you please lose the `gh::` prefix, for consistency? Thanks.

statnmap: Absolutely. Done.

gaborcsardi:
Thanks!

PR #174, opened 2023-02-06
Switch from httr to httr2

### LINKED ISSUE #67: Rate limit helpers
i.e. see how many requests you have left in the current period, and another function to automatically wait until you have another request

hadley:
And add rate limiting. Fixes #67.


R/gh_request.R
```diff
@@ -182,16 +181,6 @@ gh_send_headers <- function(accept_header = NULL, headers = NULL) {
   )
 }
 
-#' @importFrom httr write_disk write_memory
```
hadley, line 185: No longer needed because handled by `req_perform(path)`

R/gh_response.R
```diff
@@ -1,33 +1,35 @@
...
     isTRUE(grepl("param=raw$", gh_media_type, ignore.case = TRUE))
-  is_ondisk <- inherits(response$content, "path")
+  is_ondisk <- inherits(resp$body, "httr2_path")
+
   if (is_ondisk) {
-    res <- response$content
+    res <- as.character(resp$body)
+  } else if (is.na(content_type)) {
+    # empty body
+    res <- list()
```
hadley, line 18: This was previously `NULL`, which I'm pretty sure is incorrect because we add attributes below.

R/gh_response.R
```diff
@@ -1,33 +1,35 @@
...
       warning("Response came back as html :(", call. = FALSE)
     }
-    res <- list(message = content(response, as = "text"))
+    res <- list(message = httr2::resp_body_string(resp))
   }
 
-  attr(res, "method") <- response$request$method
-  attr(res, "response") <- headers(response)
-  attr(res, ".send_headers") <- response$request$headers
+  # attr(res, "method") <-  response$request$method
```
hadley, line 30: I can bring these back if you think they're important, but nothing tests them currently.

tests/testthat/test-mock-repos.R
```diff
@@ -7,113 +7,50 @@ test_that("repos, some basics", {
...
-  )
-  expect_equal(res$name, test_repo)
-  expect_equal(res$description, "Test repo for gh")
-  expect_equal(res$homepage, "https://github.com/r-lib/gh")
-  expect_false(res$private)
-  expect_false(res$has_issues)
-  expect_false(res$has_wiki)
-
-  ## TODO: POST /orgs/{org}/repos
+test_that("can POST, PATCH, and DELETE", {
```
hadley, line 27: My intuition is that these tests are really for the different verbs so I switched from modifying repos to modifying gists, which is much less invasive.

hadley:
@gaborcsardi if we decided to go down this path, I'll consider testing more broadly, doing some more refactoring and adding tests to get better coverage. But I have invested relatively little time at this point, so it would be easier enough to give it up.

A few reasons why it'd be worthwhile to switch to httr2:

* Adding retry with delay is a single line
* `last_response()` and `last_request()` make debugging in downstream packages much easier
* Could likely do further refactoring to use `req_template()` (but I think that'll require some httr2 changes because of the way that gh puts unmatched args in the body)

R/gh_response.R
gaborcsardi: Good point. OTOH I am afraid that some packages might rely on this being `NULL`, in particular usethis.

R/gh_response.R
gaborcsardi: I agree that probably only the response headers are important. OTOH, we can keep things as they were, just to be on the safe side.

gaborcsardi (commented):
I am fine with switching to httr2, it also helps httr2 to get more usage.
Looks good, I left some comments. In particular, we could make sure that returning `list()` for an empty response does not break anything in usethis.

Some tests cases are failing for me, I _think_ these might be fragile.
 
```
Failure (test-gh_whoami.R:10): whoami works in presence of PAT
res\[\["scopes"\]\] does not match "\\buser\\b".
Actual value: "delete:packages, delete_repo, read:org, repo, workflow, write:packages"
Backtrace:
 1. testthat::expect_match(res[["scopes"]], "\\buser\\b")
      at test-gh_whoami.R:10:2
 2. testthat:::expect_match_(...)
```

```
Error (test-mock-repos.R:33): can POST, PATCH, and DELETE
<github_error/http_error_404/rlang_error/error/condition>
Error in `gh_process_response(raw)`: GitHub API error (404): Not Found
x URL not found: <https://api.github.com/gists>
i Read more at <https://docs.github.com/rest/reference/gists#create-a-gist>
Backtrace:
 1. gh::gh(...)
      at test-mock-repos.R:33:2
 2. gh:::gh_process_response(raw)
      at gh/R/gh.R:176:2
```

```
Error (test-mock-repos.R:33): can POST, PATCH, and DELETE
<github_error/http_error_404/rlang_error/error/condition>
Error in `gh_process_response(raw)`: GitHub API error (404): Not Found
x URL not found: <https://api.github.com/gists>
i Read more at <https://docs.github.com/rest/reference/gists#create-a-gist>
Backtrace:
 1. gh::gh(...)
      at test-mock-repos.R:33:2
 2. gh:::gh_process_response(raw)
      at gh/R/gh.R:176:2
```

R/gh_response.R
hadley: Hmmm I thought setting an attribute on `NULL` coerced it to a list, but instead it errors:

``` r
x <- NULL
attr(x, "boo") <- 1
#> Error in attr(x, "boo") <- 1: attempt to set an attribute on NULL
```

<sup>Created on 2023-02-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

So I don't understand why this code currently works.

gaborcsardi: Oh, OK, then I must have remembered something else.

DESCRIPTION
```diff
@@ -15,10 +15,10 @@ Depends: R (>= 3.4)
 Imports:
     cli (>= 3.0.1),
     gitcreds,
-    httr (>= 1.2),
+    httr2,
     ini,
     jsonlite,
-    rlang
+    rlang (>= 1.0.0)
```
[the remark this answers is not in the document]
hadley: Let's just import it, since that's our modern standard.

R/gh_response.R
hadley: All the usethis test pass with this version of gh and I've inspect the usethis code for calls to `DELETE`, and it looks ok so. I'll see if I can come up with a test for a GET call that doesn't return any output (maybe a search for something that doesn't exist) and I'll also installed locally so I can see if anything comes up in interactive use.

gaborcsardi: I think maybe the `PUT`, `POST`, etc. requests return an empty response, no? 

Usethis has some "manual" tests, those are the ones that gh usually breaks. But we can ask Jenny to run them once we are done with the transition to httr2, before the next gh release.

hadley: I've found three cases that all return empty bodies and added them to the tests.

R/gh_response.R
hadley: Slightly hacky, but now restored.

hadley:
Hmmm, those tests assume that you have a PAT created with our default scopes. I can skip if the scopes aren't as expected, if you want.

R/gh.R
```diff
@@ -234,21 +233,84 @@ gh_response_length <- function(res) {
...
+  if (is.raw(x$body)) {
+    req <- httr2::req_body_raw(req, x$body)
+  } else {
+    req <- httr2::req_body_json(req, x$body, null = "list", digits = 4)
+  }
+  req <- httr2::req_headers(req, !!!x$headers)
+
+  # use retry-after info when possible
+  # https://docs.github.com/en/rest/overview/resources-in-the-rest-api#exceeding-the-rate-limit
+  github_is_transient <- function(resp) {
```
gaborcsardi, line 253: Should we make this behavior configurable? In some situations, I don't want to wait an hour.

hadley: Yeah, I wondered about that. And maybe we need some support for https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#secondary-rate-limits too?

R/gh.R
```diff
@@ -65,7 +65,7 @@
 #' @param .params Additional list of parameters to append to `...`.
 #'   It is easier to use this than `...` if you have your parameters in
 #'   a list already.
-#'
+#' @param .max_wait Maximum number of minutes to wait if rate limited.
```
hadley, line 68: @gaborcsardi what do you think of this name, units (i.e. minutes not seconds), and default value?

gaborcsardi: I agree that people usually want minutes, but we usually use seconds, so I would use seconds here as well. But it is not a very strong opinion.

R/gh_rate_limit.R
```diff
@@ -20,6 +20,11 @@
 
 gh_rate_limit <- function(response = NULL, .token = NULL, .api_url = NULL, .send_headers = NULL) {
   if (is.null(response)) {
+    lifecycle::deprecate_warn(
```
gaborcsardi, line 23: I think we can use `deprecate_soft()`, and also, is there a way to only emit a message instead of a warning? I don't think we need to remove this function, it is effectively  zero maintenance, and it can potentially break people's workflows. But the new function is better, so we could point people to that.

hadley: FWIW github has deprecated that the `rate` component of the `rate_limit` API: https://docs.github.com/en/rest/rate-limit?apiVersion=2022-11-28 and I don't think there's much cost in forwarding people to it. But I don't feel that strongly and I could just remove the deprecation if you don't like it.

gaborcsardi: In that case we could just take the info from `resources$core` instead, I guess. And switch to `deprecate_soft()`?

hadley: It's not that easy to make that work because the return value actually quite different — `gh_rate_limits()` gives you a run down of all the rate limits used by GH in a data frame where `gh_rate_limit()` gives you the specifics for a request with a list.

gaborcsardi: Or maybe make it superseded, and use `signal_stage()`?

gaborcsardi: In that case we can keep it as is, and make it superseded? I doubt that GH will remove it any time soon.

hadley: I don't really care either, so I'll just leave as is for now.

PR #175, opened 2023-02-14
Implement request_throttling

hadley:
Testing this revealed that I failed to set up `max_wait` correctly; it got discarded by the final line of `gh_build_request()`. And even if I had caught that, it still wasn't recorded in the response object so pagination could preserve the value.

So now the request captures the complete request list made up by gh and reuses that when paginating. I'll probably refactor this in the future to reuse the same httr2 request.

### DIFF
--- a/R/gh.R
+++ b/R/gh.R
@@ -67,6 +67,8 @@
 #'   a list already.
 #' @param .max_wait Maximum number of seconds to wait if rate limited.
 #'   Defaults to 10 minutes.
+#' @param .max_rate Maximum request rate in requests per second. Set
+#'   this to automatically throttle requests.
 #' @return Answer from the API as a `gh_response` object, which is also a
 #'   `list`. Failed requests will generate an R error. Requests that
 #'   generate a raw response will return a raw vector.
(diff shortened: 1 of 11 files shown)


R/gh.R
```diff
@@ -273,6 +276,10 @@ gh_make_request <- function(x, error_call = caller_env()) {
     )
   }
 
+  if (!is.null(x$max_rate)) {
+    req <- httr2::req_throttle(req, x$max_rate)
```
gaborcsardi, line 280: Should we make an attempt to create some meaningful `realm` here? Otherwise packages might interfere.

hadley: The realm is automatically taken from the hostname, which I think should be correct.

gaborcsardi: Right!

R/gh_response.R
```diff
@@ -24,9 +24,8 @@ gh_process_response <- function(resp) {
     res <- list(message = httr2::resp_body_string(resp))
   }
 
-  attr(res, "method") <- resp$method
   attr(res, "response") <- httr2::resp_headers(resp)
-  attr(res, ".send_headers") <- httr2::last_request()$headers
+  attr(res, "request") <- gh_req
```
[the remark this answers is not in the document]
hadley: Added some tests to hopefully prevent me from continuing to remove it 😄

PR #195, opened 2024-04-19
docs: refer to {gitcreds} more consistently

### LINKED ISSUE #194: Recommend gitcreds more?
I was reading the docs/code to find where the fallback on the `GITHUB_TOKEN` environment variable happens (in gitcreds :sweat_smile:) and noticed that while gitcreds docs state

```r
#' However, we still suggest that you add your token to the git credential
#' store with `gitcreds_set()` and remove `GITHUB_PAT` from your
#' `.Renviron` file. The credential store is more secure than storing
#' tokens in files, and command line git also uses the credential store
#' for password-less authentication.
```

here in gh,

- the README doesn't mention gitcreds at all.
- the docs of the param .token are

https://github.com/r-lib/gh/blob/1ac24e436647c608b145e848ea994fe99c8ffe8f/R/gh.R#L43-L46

- the example mention the environment variables https://github.com/r-lib/gh/blob/1ac24e436647c608b145e848ea994fe99c8ffe8f/R/gh.R#L94

maelle:
Fix #194

### DIFF
--- a/R/gh.R
+++ b/R/gh.R
@@ -40,10 +40,7 @@
 #' @param .overwrite If `.destfile` is provided, whether to overwrite an
 #'   existing file.  Defaults to `FALSE`. If an error happens the original
 #'   file is kept.
-#' @param .token Authentication token. Defaults to `GITHUB_PAT` or
-#'   `GITHUB_TOKEN` environment variables, in this order if any is set.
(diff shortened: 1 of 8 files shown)


gaborcsardi:
Thank you!

PR #200, opened 2024-08-15
Document that PATs can now expire w/ add default expiration

RodDalBen:
Document that PATs can now expire and the default expiration time is 30 days #160


lionel-:
Thanks!

PR #201, opened 2024-08-15
add vctrs methods to make rectangling nicer

### LINKED ISSUE #161: Make `gh_response` play more nicely with vctrs / tidyr's rectangling toolkit
I've been doing lots of API work lately with the `master` --> `main` task. And I've used the rectangling toolkit from tidy a lot. As it stands, I have to `unclass()` lists of gh responses everywhere, like so:

``` r
raw_pr_dat <- raw_prs %>%
  # strip the gh_response class to make tidyr happy about unnesting
  map(unclass) %>%
  enframe(name = "spec", value = "raw_prs") %>%
  rowwise() %>%
  filter(length(raw_prs) > 0) %>%
  unnest_longer(raw_prs)
```

I think the class should be `c("gh_response", "list")`. I got confirmation from @DavisVaughan that has been thought about over in vctrs and this is the recommended solution.

tanho63:
Closes #161 

Adds vctrs methods provided by @DavisVaughan to help `gh_response` play more nicely with vctrs rectangling toolkit. 

test instructions:

```r
pkgload::load_all()
x <- gh("/users", .limit = 2)
Sys.sleep(1)
y <- gh("/users", .limit = 2)

str(vctrs::vec_c(x, y), list.len = 3)

list(x, y) |> 
  tibble::tibble() |> 
  tidyr::unnest_longer(1)
```


hadley:
Could you include a simple test case too? I think using `vec_c()` should be sufficient?

tests/testthat/test-gh_response.R
```diff
@@ -83,3 +83,14 @@ test_that("output file is not overwritten on error", {
...
+
+test_that("gh_response objects can be combined via vctrs #161",{
+  skip_on_cran()
+  skip_if_not_installed("vctrs")
+  user_1 <- gh("/users", .limit = 1)
+  user_2 <- gh("/users", .limit = 1,)
+  user_vec <- vctrs::vec_c(user_1, user_2)
+  user_df <- vctrs::vec_rbind(user_1[[1]], user_2[[1]])
+  expect_equal(length(user_vec), 2)
+  expect_equal(nrow(user_df), 2)
```
tanho63, line 95: decided that vec_rbind was slightly different but probably closer to the desired use case than vec_c (and therefore worth testing separately) - could also keep only the vec_rbind test since it's closer to the original rectangling issue, if preferred?

hadley: The existing approach seems fine to me.

PR #202, opened 2024-08-15
refactor gh::gh parameter handling to make testable without mocking

### LINKED ISSUE #197: Use `testthat::with_mock()` instead of mockery

tanho63:
Closes #197. 

Original issue suggested replacing mockery with `testthat::with_mock`, however the usage of mockery was to test parameter handling (accepting via both ... and a .params arg) so naive replacement caused an infinitely recursive call to gh_build_request. 

Resolved by refactoring parameter handling into its own function, `.parse_params()`, which is easier to test conventionally

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -76,6 +76,12 @@ drop_named_nulls <- function(x) {
   cleanse_names(x[!named | !null])
 }
 
+.parse_params <- function(..., .params = list()) {
+  params <- c(list(...), .params)
+  params <- drop_named_nulls(params)
+  return(params)
+}
+
 check_named_nas <- function(x) {
   if (has_no_names(x)) {
     return(x)
--- a/R/utils.R
+++ b/R/utils.R
@@ -78,8 +78,7 @@ drop_named_nulls <- function(x) {
 
 .parse_params <- function(..., .params = list()) {
   params <- c(list(...), .params)
-  params <- drop_named_nulls(params)
-  return(params)
+  drop_named_nulls(params)
 }
 
 check_named_nas <- function(x) {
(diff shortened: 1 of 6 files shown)


tests/testthat/test-utils.R
```diff
@@ -67,3 +67,15 @@ test_that("named NA is error", {
...
+
+test_that(".parse_params combines list .params with ... params", {
+  params <- list(
+    .parse_params(org = "ORG", repo = "REPO", number = "1"),
+    .parse_params(org = "ORG", repo = "REPO", .params = list(number = "1")),
+    .parse_params(.params = list(org = "ORG", repo = "REPO", number = "1"))
+  )
+
+  expect_identical(params[[1]], params[[2]])
+  expect_identical(params[[2]], params[[3]])
```
lionel-, line 80: I think the tests would be clearer if you assigned the results to named objects, perhaps: `params`, `params_partial_list`, and `params_list`?

R/utils.R
```diff
@@ -76,6 +76,12 @@ drop_named_nulls <- function(x) {
   cleanse_names(x[!named | !null])
 }
 
+.parse_params <- function(..., .params = list()) {
+  params <- c(list(...), .params)
+  params <- drop_named_nulls(params)
+  return(params)
```
lionel-, line 82: nitpick: Unnecessary `return()`

```suggestion
  drop_named_nulls(params)
```

tests/testthat/test-utils.R
lionel-: But I see that's just following what existed before so let's just merge.

lionel- (approved):
That's quite an improvement! Just two nitpicks.

PR #211, opened 2024-09-07
Reduce connection timeout to 5s

hadley:


gaborcsardi:
@hadley I think this is not the commit that you wanted in this PR?

PR #212, opened 2024-09-07
New `gh_token_exists()`

hadley:
I think this will be useful for improving some error messages in usethis.


gaborcsardi (approved):
Thanks!

R/gh_token.R
```diff
@@ -53,6 +53,20 @@ gh_token <- function(api_url = NULL) {
...
+gh_token_exists <- function(api_url = NULL) {
+  api_url <- api_url %||% default_api_url()
+  tryCatch(
+    {
+      token <- gitcreds::gitcreds_get(get_hosturl(api_url))
+      gh_pat(token$password)
+      TRUE
+    } ,
+    error = function(e) FALSE
+  )
```
jennybc, line 67: Should this be factored in such a way that it can be used inside `gh_token()` just above?

jennybc (commented):
> I think this will be useful for improving some error messages in usethis.

Just for context usethis does have a lot of existing logic re: detecting a PAT and identifying any problems with it (most important ones: exists but has expired, exists but is under-scoped).

https://github.com/r-lib/usethis/blob/2cc9e5a4e5e3034399fafb65d871eec352d83a15/R/github_token.R#L122-L220

R/gh_token.R
hadley: It's not immediately obvious to me how to do that due to a few subtle differences in the functions.

PR #213, opened 2024-09-09
Reduce connection timeout to 5s

hadley:
This makes gh fail a little faster in the case that GitHub is not available (e.g. you're on bad airplane wifi)


gaborcsardi:
Thanks!

PR #217, opened 2024-11-28
Add support for viewer-based credentials on Posit Connect

atheriel:
This commit brings support for Posit Connect's ["viewer-based credentials" feature][0] to `gh`.

~~Similar to [the recent work in `odbc`][1], the way this is exposed to R users is to require them to pass a Shiny session argument to `gh_token()`, either directly or via `gh()` or `gh_gql()`.~~

Checks for viewer-based credentials are designed to fall back gracefully to existing authentication methods. This is intended to allow users to -- for example -- develop and test a Shiny app that uses GitHub credentials in desktop Positron/RStudio or Posit Workbench and deploy it with no code changes to Connect.

Most of the actual work is outsourced to a new shared package, [`connectcreds`][2].

Unit tests are included.

[0]: https://docs.posit.co/connect/user/oauth-integrations/
[1]: https://github.com/r-dbi/odbc/pull/853
[2]: https://github.com/posit-dev/connectcreds/

### DIFF
--- a/R/gh_token.R
+++ b/R/gh_token.R
@@ -46,8 +46,16 @@
 gh_token <- function(api_url = NULL) {
   api_url <- api_url %||% default_api_url()
   stopifnot(is.character(api_url), length(api_url) == 1)
+  host_url <- get_hosturl(api_url)
+  # Check for credentials supplied by Posit Connect.
+  if (is_installed("connectcreds")) {
+    if (connectcreds::has_viewer_token(host_url)) {
+      token <- connectcreds::connect_viewer_token(host_url)
+      return(gh_pat(token$access_token))
+    }
+  }
   token <- tryCatch(
-    gitcreds::gitcreds_get(get_hosturl(api_url)),
+    gitcreds::gitcreds_get(host_url),
     error = function(e) NULL
   )
   gh_pat(token$password %||% "")
@@ -56,15 +64,7 @@ gh_token <- function(api_url = NULL) {
 #' @export
 #' @rdname gh_token
 gh_token_exists <- function(api_url = NULL) {
-  api_url <- api_url %||% default_api_url()
-  tryCatch(
-    {
-      token <- gitcreds::gitcreds_get(get_hosturl(api_url))
-      gh_pat(token$password)
-      TRUE
-    } ,
-    error = function(e) FALSE
-  )
+  tryCatch(nzchar(gh_token(api_url)), error = function(e) FALSE)
 }
 
 gh_auth <- function(token) {
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -23,6 +23,7 @@ Imports:
     lifecycle,
     rlang (>= 1.0.0)
 Suggests:
+    connectcreds,
     covr,
     knitr,
     rmarkdown,
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,9 @@
 * `gh()` now uses a cache provided by httr2. This cache lives in `tools::R_user_dir("gh", "cache")`, maxes out at 100 MB, and can be disabled by setting `options(gh_cache = FALSE)` (#203).
 * Removes usage of mockery (@tanho63, #197)
 
+* `gh_token()` can now pick up on the viewer's GitHub credentials (if any) when
+  running on Posit Connect (@atheriel, #217).
+
 # gh 1.4.1
 
 * `gh_next()`, `gh_prev()`, `gh_first()` and `gh_last()`
--- a/tests/testthat/test-gh_token.R
+++ b/tests/testthat/test-gh_token.R
@@ -163,3 +163,11 @@ test_that("get_apiurl() works", {
   expect_equal(get_apiurl("https://github.acme.com/OWNER/REPO"), x)
   expect_equal(get_apiurl("https://github.acme.com/api/v3"), x)
 })
+
+test_that("tokens can be requested from a Connect server", {
+  skip_if_not_installed("connectcreds")
+
+  token <- strrep("a", 40)
+  connectcreds::local_mocked_connect_responses(token = token)
+  expect_equal(gh_token(), gh_pat(token))
+})


atheriel:
This is an edited version of the example from Connect's cookbook that shows how this works in practice:

```r
library(shiny)
library(bslib)
library(gh)

ui <- page_sidebar(
  title = "My GitHub Issues",
  sidebar = sidebar(
    title = "GitHub Repo",
    textInput(
      "gh_repo",
      "GitHub Repository",
      value = "octocat/Hello-World",
      placeholder = "octocat/Hello-World"
    ),
  ),
  layout_column_wrap(
    card(card_header("assigned"), verbatimTextOutput("assigned")),
    card(card_header("issues"), verbatimTextOutput("issues")),
    card(card_header("prs"), verbatimTextOutput("prs")),
    card(card_header("whoami"), verbatimTextOutput("whoami"))
  )
)

server <- function(input, output, session) {
  output$assigned <- renderText ({
    assigned <- gh("GET /issues", .session = session)
    paste(assigned)
  })

  output$issues <- renderText({
    req(input$gh_repo)
    my_issues <- gh(
      "GET /repos/{gh_repo}/issues",
      gh_repo = input$gh_repo,
      .session = session
    )
    paste(my_issues)
  })

  output$prs <- renderText({
    req(input$gh_repo)
    my_pulls <- gh(
      "GET /repos/{gh_repo}/pulls",
      gh_repo = input$gh_repo,
      .session = session
    )
    paste(my_pulls)
  })

  output$whoami <- renderText({
    resp <- gh_whoami(.session = session)
    paste(resp$name, resp$login, collapse = "\n")
  })
}

shinyApp(ui, server)
```

atheriel:
Took a second go at this per @jennybc's advice; it now makes far less invasive changes to `gh`.

R/gh_token.R
```diff
@@ -142,3 +144,30 @@ obfuscate <- function(x, first = 4, last = 4) {
...
+  if (!isNamespaceLoaded("shiny")) {
+    return(NULL)
+  }
+  if (is_installed("connectcreds")) {
+    # Avoid taking a Suggests dependency on Shiny, which is otherwise
+    # irrelevant to gh.
+    f <- get("getDefaultReactiveDomain", envir = asNamespace("shiny"))
+    return(f())
+  }
+  cli::cli_inform(
```
nealrichardson, line 161: If this is running in Connect, where do you expect someone to see this message?

This also seems to assume that if I'm running on Connect, then I must want to use viewer credentials. What if I've put a PAT in an encrypted env var?

R/gh_token.R
```diff
@@ -142,3 +144,30 @@ obfuscate <- function(x, first = 4, last = 4) {
     substr(x, start = nchar(x) - last + 1, stop = nchar(x))
   )
 }
+
+get_connect_session <- function() {
```
nealrichardson, line 148: This is really the Shiny session, when running on Connect, right?

Might be relevant to make that distinction because with the client credentials flow, IIUC it will be possible to use oauth tokens from Connect in non-Shiny contexts (R markdown or otherwise).

R/gh_token.R
atheriel: It shows up in Connect's logs... which is not ideal, but it's all I can think of. Here is a screenshot of an earlier version of the message:

Screenshot from 2024-12-05 14-49-41

And it's intentionally an `info` suggestion here, not a warning or error, because I absolutely expect that publishers will want to use PATs instead for some apps, too.

R/gh_token.R
atheriel: Yes, although GitHub doesn't support the client credentials flow, so this isn't directly relevant to the `gh` package.

(I'm planning to add a dedicated API to `connectcreds` for service accounts, but it won't consume a Shiny session.)

R/gh_token.R
```diff
@@ -46,8 +46,17 @@
 gh_token <- function(api_url = NULL) {
   api_url <- api_url %||% default_api_url()
   stopifnot(is.character(api_url), length(api_url) == 1)
+  host_url <- get_hosturl(api_url)
+  # Session credentials take precedence over static credentials.
+  session <- get_connect_session()
+  if (!is.null(session)) {
```
hadley, line 52: It took me a minute to realise that we don't need an explicit check for connectcreds here because that occurs in `get_connect_session()`. But I suspect that suggests these functions need to be factored slightly differently?

Maybe more like:

```R
if (has_connectcreds() && in_shinyapp()) { 
  session <- get_connect_session()
  token <- connectcreds::connect_viewer_token(session, host_url)
  if (!is.null(token)) {
    return(gh_pat(token$access_token))
  }
}
```

?

hadley (commented):
The approach of auto-detecting the session seems reasonable to me.

R/gh_token.R
atheriel: The current form is largely because I'm struggling with how error propagation should happen here.

Right now `connect_viewer_token()` throws errors while `has_viewer_token()` does not.

The trouble is that we can't assume that just because a Shiny app is deployed to Connect the publisher wants to use viewer-based credentials. So clearly it can't be fatal if there are no credentials available, and that's why I'm using `has_viewer_token()`. But at the same time, it feels bad to swallow errors like "no OAuth integration is assigned to this app" that would probably help authors debug why their app isn't working.

Currently I have a `connectcreds_debug` option that works like `gargle`'s "debug" mechanism, emitting the errors as messages instead. But it's not on by default, so folks would probably only learn about it through the documentation... limiting its usefulness. Maybe `connectcreds` should log debug messages by default if it detects we're on Connect?

Do you have any thoughts on error reporting?

hadley: It feels like we need some explicit signal from the user that they want to use viewer-based auth when deployed?

atheriel: We had that when we had an explicit `session` parameter :disappointed:.

Maybe some of this is just because `gh_token()` doesn't issue an error when there are no credentials. In `odbc` and `elmer` I notice that we can say things like "No credentials available ... by the way, you're running in Connect? Try turning on viewer-based credentials, see _link_".

hadley: So maybe we're better off switching back to that UI? 

Maybe `gh_token()` should error? But even if that is the correct behaviour there likely to be a bunch of downstream changes because of that.

atheriel:
Updated to use the latest version of `connectcreds`, which has a much simpler API, and is [available on CRAN as of this morning](https://cran.r-project.org/web/packages/connectcreds/index.html).

Because the API of that package has been dramatically simplified, the changes to `gh` are far less invasive than in previous patches. The unit test story is also much better, because `connectcreds` includes comprehensive mocking features.

R/gh_token.R
atheriel: Closing, this is no longer relevant due to the changes in the `connecreds` API.

R/gh_token.R
atheriel: Resolving; this is no longer relevant due to changes in the `connecreds` API.

gaborcsardi (approved):

gaborcsardi:
Thanks!

PR #220, opened 2025-03-29
add proxy support

Yunuuuu:
Fix https://github.com/r-lib/gh/issues/221. The title summarizes the aim of the PR.

### DIFF
--- a/R/gh.R
+++ b/R/gh.R
@@ -46,6 +46,10 @@
 #'   environment variable if set.
 #' @param .method HTTP method to use if not explicitly supplied in the
 #'    `endpoint`.
+#' @param .proxy_url,.proxy_port Location of proxy.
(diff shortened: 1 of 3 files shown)


gaborcsardi:
Thanks! But don't the libcurl proxy environment variables work already?

Yunuuuu:
Since my network is heavily restricted due to my country's policies, I want to make `usethis` git-related functions work, which depend on `gh` and `gert`.

image

Having support for proxy environment variables in `gh` would be extremely useful, as `usethis` does not expose many arguments from `gh` or `gert`, making it difficult to configure network access directly. 

I would really appreciate any advice on how to test whether the proxy environment variable is working correctly in `gh`. I couldn’t find any verbose argument for this—does one exist? Thanks in advance for your help!

gaborcsardi:
See e.g. https://everything.curl.dev/usingcurl/proxies/env.html

Issue #231, opened 2026-05-22
Length of ghs tokens increasing


jharmon-gilead:
GitHub has [started to roll out app tokens that are longer](https://github.blog/changelog/2026-04-24-notice-about-upcoming-new-format-for-github-app-installation-tokens/) and break the `validate_gh_pat()` check.

https://github.com/r-lib/gh/blob/e9194ae1bb2993ba8f72d9da2f07679b5e3e013a/R/gh_token.R#L93-L116

They say "The overall length of the tokens will be longer (~520 characters) and will vary based on the data stored within it." The "~" there is annoying, so I'm not sure EXACTLY what should be allowed, but something along these lines should work:

```r
  if (
    x == "" ||
      # https://github.blog/changelog/2021-03-04-authentication-token-format-updates/
      # Fine grained tokens start with "github_pat_".
      # https://github.blog/changelog/2022-10-18-introducing-fine-grained-personal-access-tokens/
      grepl(
        "^(gh[pousr]_[A-Za-z0-9_]{36,251}|github_pat_[A-Za-z0-9_]{36,244}|ghs_[A-Za-z0-9_]+)$",
        x
      ) ||
      grepl("^[[:xdigit:]]{40}$", x)
  ) {
    x
  } else {
    url <- "https://gh.r-lib.org/articles/managing-personal-access-tokens.html"
    cli::cli_abort(c(
      "Invalid GitHub PAT format",
      "i" = "A GitHub PAT must have one of four forms:",
      "*" = "40 hexadecimal digits (older PATs)",
      "*" = "A 'ghp_' prefix followed by 36 to 251 more characters (newer PATs)",
      "*" = "A `ghs_` prefix followed by about 500 characters (GitHub App installation tokens)",
      "*" = "A 'github_pat_' prefix followed by 36 to 244 more characters (fine-grained PATs)",
      "i" = "Read more at {.url {url}}."
    ))
  }
}
```

I'm seeing sporadic failures on GitHub Actions that look like they are caused by this.

I'll submit a PR shortly.

jeroen:
FYI https://github.blog/changelog/2026-04-24-notice-about-upcoming-new-format-for-github-app-installation-tokens/ suggest we remove the validation alltogether:


>  It’s recommended that you treat tokens as opaque strings and avoid validating them against hardcoded patterns.
> To help prepare for this change, ensure that:
>
>  - Your apps do not take a dependency on access tokens being a certain length.
[...]

PR #233, opened 2026-05-26
Only warn for unrecognized token formats

gaborcsardi:
GitHub PAT format validation now issues a warning by default instead of throwing an error, so a PAT in an unrecognized (e.g. newly introduced) format is still used. Set the `gh_validate_tokens` option or the `GH_VALIDATE_TOKENS` environment to `"off"`, `"warn"` or `"error"` to configure this.

Related to #231.


jeroen:
Thanks, lgtm

jennybc (approved):

PR #237, opened 2026-05-27
Return partial results in the condition on interrupt

### LINKED ISSUE #98: FR: Support returning incomplete results after an interrupt
Scenario: I'm running a query and didn't realize how many results are returned. The progress bar (#26) indicates that it's going to be too long, but I'd like to look at partial results. I'm pressing Escape (or Ctrl + C) but the intermediate results are gone.

As discussed in #86, we could store intermediate results in a private environment and expose with a function, e.g. `gh_partial()` . Catching interrupts is a bad idea.

gaborcsardi:
Closes #98.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,12 @@
 # gh (development version)
 
+* `gh()` now signals a classed `gh_interrupt` interrupt when a paginated
+  call is interrupted (e.g. via `Ctrl+C` / `Escape`). The condition object
+  carries the records fetched so far on its `$gh_result` field. If the
+  interrupt is not caught, then you can also access it via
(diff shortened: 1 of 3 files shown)


hadley:
Did you consider just switching to `req_perform_iterative()` so you could inherit from httr2?

gaborcsardi:
Yeah, I looked at that solution. But it does not seem like a good idea to me to ignore interrupts and return partial results downstream. I think if the user presses an interrupt in an interactive session or sends a SIGINT to a worker subprocess then they want the computation/worker stopped. This is essentially the same concern you had in the issue, hence your top-level restriction suggestion. I think this PR essentially implements that at the top level and has more predictable behavior, i.e. an interrupt will jump to top level.

